PR #4, opened 2018-10-17
Upload broken symlinks fixed

RobiNino:
Fixed jfrog-cli-go #229.

Added the ability to upload broken symlinks with --symlinks=true.

Fixed an issue where when --symlinks=false, the symlink target would be uploaded given the symlink name instead of the target name.

### DIFF
--- a/artifactory/services/upload.go
+++ b/artifactory/services/upload.go
@@ -147,15 +147,19 @@ func collectFilesForUpload(uploadParams UploadParams, producer parallel.Runner, 
 		return
 	}
 
-	isDir, err := fileutils.IsDir(uploadParams.IsSymlink(), rootPath)
+	isDir, err := fileutils.IsDirExists(rootPath, uploadParams.IsSymlink())
 	if err != nil {
 		errorsQueue.AddError(err)
 		return
 	}
 
-	// If the path is a single file return it or it is a link and preserve symbolic links is set to true
-	if !isDir || (uploadParams.IsSymlink() && fileutils.IsPathSymlink(rootPath)) {
-		artifact := fspatterns.GetSingleFileToUpload(rootPath, uploadParams.GetTarget(), uploadParams.IsFlat(), uploadParams.IsSymlink())
+	// If the path is a single file (or a symlink while preserving symlinks) upload it and return
+	if !isDir || (fileutils.IsPathSymlink(rootPath) && uploadParams.IsSymlink()) {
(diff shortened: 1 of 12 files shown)


artifactory/services/upload.go
```diff
@@ -238,12 +239,13 @@ func createUploadTask(taskData *uploadTaskData) error {
 		taskData.target = strings.Replace(taskData.target, "{"+strconv.Itoa(i)+"}", group, -1)
 	}
 	var task parallel.TaskFunc
-	taskData.target = getUploadTarget(taskData.uploadParams.IsFlat(), taskData.path, taskData.target)
-	// If case taskData.path is a symlink we get the symlink link path.
+
 	symlinkPath, e := fspatterns.GetFileSymlinkPath(taskData.path)
```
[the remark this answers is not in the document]
RobiNino: Do we need if .isSymlink is false?
Yes, because it constructs the final target path while also taking .isFlat() into account.

artifactory/services/upload.go
```diff
@@ -112,9 +112,10 @@ func addSymlinkProps(artifact clientutils.Artifact, uploadParams UploadParams) (
...
-			return "", err
-		}
-		if !fileInfo.IsDir() {
+			if !os.IsNotExist(err) { // If error occurred, but not due to nonexistence of Symlink target -> return empty
+				return "", err
+			}
+		} else if !fileInfo.IsDir() { // If Symlink target exists -> get SHA1 if isn't a directory
 			file, err := os.Open(artifact.LocalPath)
 			errorutils.CheckError(err)
 			if err != nil {
```
eyalbe4: The above error check can be simplified to:
```
			if err != nil {
				return "", CheckError(err)
			}
```
and as a result, the following line can be removed:
```
			errorutils.CheckError(err)
```

utils/io/fileutils/files.go
```diff
@@ -27,20 +27,51 @@ func GetFileSeparator() string {
...
+		}
 		return false, errorutils.CheckError(err)
 	}
-	return !f.IsDir(), nil
+	return fileInfo.IsDir(), nil
+}
+
+// Get the file info of the file in path.
+// If path points at a symlink and `preserveSymLink == true`, return the file info of the symlink instead
+func GetFileInfoByPreserveSymLink(path string, preserveSymLink bool) (fileInfo os.FileInfo, err error) {
```
eyalbe4: I suggest shortening this function name to  "GetFileInfo".

eyalbe4:
Added a few more inline comments @RobiNino.

PR #9, opened 2018-11-07
Get Artifactory version from body

AlexeiVainshtein:


artifactory/auth/rtdetails.go
```diff
@@ -122,10 +123,14 @@ func (rt *artifactoryDetails) getArtifactoryVersion() (string, error) {
...
+	var version ArtifactoryVersion
+	err = json.Unmarshal(body, &version)
+	if err != nil {
+		return "", errorutils.CheckError(err)
 	}
-	return strings.TrimSpace(serverValues[1]), errorutils.CheckError(err)
+	return strings.TrimSpace(version.Version), nil
+}
+
+type ArtifactoryVersion struct {
```
eyalbe4: Can the *ArtifactoryVersion* type be made private and renamed to *artifactoryVersion*?

eyalbe4:
Looks good @AlexeiVainshtein.
See my inline question.
Also, I suggest adding an integration test for the GetVersion API.

artifactory/auth/rtdetails.go
AlexeiVainshtein: Yes we can. I did the change and will do a PR

PR #11, opened 2018-11-12
SSH Authentication behaviour update

RobiNino:
Following Issue 83 on the CLI, the behaviour of ssh authentication was changed to:
1. Try authenticating via ssh-agent.
If failed due to any reason (No agent / connection failed / no key in agent):
2. Try authenticating via ssh-key.
If the key is encrypted, passphrase has to be passed via the --ssh-passphrase flag.

### DIFF
--- a/artifactory/auth/sshlogin.go
+++ b/artifactory/auth/sshlogin.go
@@ -3,33 +3,72 @@ package auth
 import (
 	"bytes"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
+	"fmt"
 	"github.com/jfrog/jfrog-client-go/utils"
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 	"github.com/jfrog/jfrog-client-go/utils/log"
+	"github.com/jfrog/jfrog-client-go/utils/prompt"
 	"github.com/xanzy/ssh-agent"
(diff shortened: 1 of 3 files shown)


artifactory/auth/sshlogin.go
```diff
@@ -65,12 +103,38 @@ func sshAuthentication(url string, sshKey, sshPassphrase []byte) (map[string]str
...
+		return nil, nil, errorutils.CheckError(err)
+	}
+	if len(sshPassphrase) == 0 {
+		encryptedKey, err := IsEncrypted(sshKey)
+		if err != nil {
+			return nil, nil, errorutils.CheckError(err)
+		}
+		// If key is encrypted but no passphrase specified
+		if encryptedKey {
+			return nil, nil, errorutils.CheckError(errors.New("SSH Key is encrypted but no passphrase was specified. Please pass a passphrase with the --ssh-passphrase flag"))
```
eyalbe4: This code is not necessarily used by JFrog CLI (it is part of jfrog-client-go, which is a library which can be used by any client).
I therefore suggest removing the second part of the error message:
**...Please pass a passphrase with the --ssh-passphrase flag**

eyalbe4:
Looks good @RobiNino!
See my inline comment.

PR #28, opened 2018-12-24
Minor improvements to search aql

RobiNino:


eyalbe4:
@RobiNino, to be on the safe side, maybe it is worth letting @barbelity take a look at this change, because he has recently worked on this area of the code.

barbelity:
Looks good to me.

PR #29, opened 2019-01-06
Add timeouts to Artifactory http-client

barbelity:


tests/utils.go
```diff
@@ -49,23 +51,43 @@ func init() {
...
 	testsSearchService.ArtDetails = getArtDetails()
 }
 
 func createArtifactoryDeleteManager() {
-	testsDeleteService = services.NewDeleteService(httpclient.NewDefaultHttpClient())
+	client, err := httpclient.ClientBuilder().Build()
+	if err != nil {
+		log.Error(fmt.Sprintf(HttpClientCreationFailureMessage, err.Error()))
+		os.Exit(1)
+	}
```
eyalbe4: The above lines are duplicated three times in this file.
I suggest we put them in a private function and reuse it.

eyalbe4:
Awesome work @barbelity!
See my inline comment.

PR #30, opened 2019-01-10
SSH token expiry handeling

RobiNino:


errors/httperrors/errors.go
```diff
@@ -36,3 +38,26 @@ func IsResponseStatusError(err error) bool {
 	_, isRespErr := err.(*ResponseStatusError)
 	return isRespErr
 }
+
+func GetStatusCodeFromResponseStatusError(err error) (int, error) {
```
eyalbe4: I suggest removing the GetStatusCodeFromResponseStatusError function, because its usage is a bit dangerous.
As a replacement for this function I suggest the following:
In all the places this function is current use, check the response status returned from the http client.
Do acheieve this, you may need to add the response as an additional returned parameter from the http client methods.

eyalbe4:
I reviewed some of the code, not all of it. Will proceed soon.

Please note that @barbelity made significant changes to the http client recently. That's the reason for the conflicts. His changes are pretty straight forward, so I think it'll be easy to adjust following a quick sync with him.

artifactory/utils/wrappers/httpclient.go
```diff
@@ -0,0 +1,251 @@
...
+func NewArtifactoryHttpClient(artDetails *auth.ArtifactoryDetails, client *http.Client) *ArtifactoryHttpClientWrapper {
+	rtHttpClient := &ArtifactoryHttpClientWrapper{}
+	rtHttpClient.ArtDetails = artDetails
+	rtHttpClient.httpClient = httpclient.NewHttpClient(client)
+	return rtHttpClient
+}
+
+func (rtc *ArtifactoryHttpClientWrapper) SendGet(url string, followRedirect bool, httpClientsDetails *httputils.HttpClientDetails) (resp *http.Response, respBody []byte, redirectUrl string, err error) {
+	// A 'do while' loop that handles a token expiry if necessary
+	for isNewToken := true; isNewToken; {
```
eyalbe4: I suggest modifying all the loops in this files, so that they do not loop more than twice.
The current implementation may not cause a bug, but I guess they'd never need to run more than twice anyway, and it is best to avoid an endless loop (due to a future bug). Better to receive an error straight away than trying to figure our what causes the code to hang. 

And this raises another question - if I'm using the wrong SSH key, would the SSH login return an error or an 401 unauthorised? If the later is returned, then couldn't we end up with the an endless loop with the current implementation?

artifactory/auth/rtdetails.go
```diff
@@ -90,27 +125,100 @@ func (rt *artifactoryDetails) SetAccessToken(accessToken string) {
...
+	// If an unauthorized ssh connection -> ssh token has expired.
+	if statusCode == http.StatusUnauthorized && rt.IsSshAuthentication() {
+		return rt.handleSshTokenExpiry(httpClientDetails)
+	}
+	return false, nil
+}
+
+// Handles the process of acquiring a new ssh token from server (if one wasn't acquired yet) and returns true.
+// Returns false if an error has occurred.
+func (rt *artifactoryDetails) handleSshTokenExpiry(httpClientDetails *httputils.HttpClientDetails) (bool, error) {
```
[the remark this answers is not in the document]
RobiNino: The function returns an error if such occurred, and so does the it's parent function. Can you point to a place in code where the error is swallowed?
The boolean value can be removed, but I think it adds readability to the code, and an 'if else' will be needed in the parent function to determine if a token has changed through the error. What do you think?

artifactory/auth/rtdetails.go
```diff
@@ -43,7 +59,11 @@ type artifactoryDetails struct {
 	ApiKey         string            `json:"-"`
 	AccessToken    string            `json:"-"`
 	version        string            `json:"-"`
+	SshUrl         string            `json:"-"`
+	SshKeyPath     string            `json:"-"`
+	SshPassphrase  string            `json:"-"`
 	SshAuthHeaders map[string]string `json:"-"`
+	TokenMutex     sync.Mutex
```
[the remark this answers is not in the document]
RobiNino: Removed

artifactory/utils/wrappers/httpclient.go
RobiNino: I will limit the loop to run twice, and throw an error if a new token wasn't acquired by then.

When using the wrong SSH key (rsa key) the initial authentication will fail, so we will not get to this point. The program will exit on error in an earlier stage in this case.
Provided that the initial authenticating was successful, if the token expired the later is returned. The HandleTokenExpiry function is responsible for acquiring a new token, which will cause the loop to break.

errors/httperrors/errors.go
RobiNino: Done

PR #31, opened 2019-01-15
Change the way Logger is created

barbelity:
Logger will no longer take the log-level from JFROG_CLI_LOG_LEVEL.
Instead, it will receive its log-level value from the log creator.

### DIFF
--- a/utils/log/logger.go
+++ b/utils/log/logger.go
@@ -19,13 +19,13 @@ const (
 
 func init() {
 	if Logger == nil {
-		Logger = NewLogger("")
+		Logger = NewLogger(INFO)
 	}
 }
 
-func NewLogger(logLevel string) Log {
+func NewLogger(logLevel LevelType) Log {
 	logger := new(jfrogLogger)
-	logger.SetLogLevel(GetCliLogLevel(logLevel))
+	logger.SetLogLevel(logLevel)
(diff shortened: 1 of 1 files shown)


utils/log/logger.go
```diff
@@ -19,18 +19,13 @@ const (
...
-func NewLogger() Log {
+func NewLogger(logLevel string) Log {
 	logger := new(jfrogLogger)
-	logLevel := os.Getenv("JFROG_CLI_LOG_LEVEL")
-	if logLevel != "" {
-		logger.SetLogLevel(GetCliLogLevel(logLevel))
-	} else {
-		logger.SetLogLevel(INFO)
-	}
+	logger.SetLogLevel(GetCliLogLevel(logLevel))
```
eyalbe4: I think we should remove the **GetCliLogLevel** function, because the SetLogLevel API can receive a log level enum.

utils/log/logger.go
```diff
@@ -19,18 +19,13 @@ const (
 
 func init() {
 	if Logger == nil {
-		Logger = NewLogger()
+		Logger = NewLogger("")
 	}
 }
 
-func NewLogger() Log {
+func NewLogger(logLevel string) Log {
```
eyalbe4: It is probably better for this function to receive a log level enum, instead of a string.

PR #32, opened 2019-01-21
Add possibility to set http client timeout.

zbindenren:
It is a bad practice to use the default http client, because it has no timeout [configured](https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779).

This pull request fixes this issue.

### DIFF
--- a/artifactory/config.go
+++ b/artifactory/config.go
@@ -1,6 +1,8 @@
 package artifactory
 
 import (
+	"time"
+
 	"github.com/jfrog/jfrog-client-go/artifactory/auth"
 	"github.com/jfrog/jfrog-client-go/utils/log"
 )
@@ -13,6 +15,7 @@ type artifactoryServicesConfig struct {
 	minSplitSize      int64
 	splitCount        int
 	minChecksumDeploy int64
+	timeout           time.Duration
 	logger            log.Log
 }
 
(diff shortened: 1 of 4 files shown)


eyalbe4:
You're right @zbindenren - thanks for this PR.
Please take a look at @barbelity's PR - https://github.com/jfrog/jfrog-client-go/pull/29
It has already been merged into the *dev* branch, and will eventually be merged into *master* and released.
We'd appreciate your feedback for the PR. Please let us know if you think this PR is still needed.

zbindenren:
@eyalbe4 sorry, I didn't see that. With that pull request you theoretically get even a more granular control for timeouts. 

What I am still missing is see [here](https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/):

```go
        ResponseHeaderTimeout: 10 * time.Second,
        ExpectContinueTimeout: 1 * time.Second,
```

And I don't see how I can customize those timeouts.

zbindenren:
And there is a typo [createDeafaultHttpTransport()](https://github.com/jfrog/jfrog-client-go/pull/29/files#diff-018ca4d91b783eaec414d24ff7533348R37).

zbindenren:
@barbelity I think this is best practice in most packages. For instance the [etcdclient](https://github.com/etcd-io/etcd/tree/master/clientv3) package let you configure a RequestTimeout and a DialTimeout. 

I personally prefer to have just one timeout option:

```go
&http.Client{
    Timeout: 15 * time.Second,
}
```

which is configurable and has a sane default than all options which I cannot touch.

barbelity:
@zbindenren we think that using the http.Client.Timeout may be problematic. For example, in case of downloading a large file from Artifactory - the suggested timeout will most probably hit and the connection will be closed before the download completes.
The values set in the client's Transport are meant to prevent cases in which the connection hangs without traffic, yet allow downloading and uploading large files.

Your comment is highly appreciated, and it is best if you could share with us a case for which the existing configuration should be improved.

zbindenren:
@barbelity how about to add the possibility to configure a the artifactory httpclient with a `http.Client`?

PR #33, opened 2019-01-23
Create RetryExecutor + add retries to client's download and upload operations

barbelity:


httpclient/client.go
```diff
@@ -154,15 +149,43 @@ func setRequestHeaders(httpClientsDetails httputils.HttpClientDetails, size int6
...
 func (jc *HttpClient) UploadFile(localPath, url string, httpClientsDetails httputils.HttpClientDetails, retries int) (resp *http.Response, body []byte, err error) {
-	for i := 0; i < retries+1; i++ {
-		resp, body, err = jc.doUploadFile(localPath, url, httpClientsDetails)
-		if resp != nil && resp.StatusCode < 500 {
-			// No error and status < 500
-			return
-		}
-		log.Warn("Upload attempt #", i, "to", url, "failed -", getFailureReason(resp, err))
+	retryExecutor := retry.RetryExecutor{
+		RetriesNum:      retries,
```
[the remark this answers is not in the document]
barbelity: Changed.

artifactory/services/readfile.go
```diff
@@ -54,5 +54,5 @@ func (ds *ReadFileService) ReadRemoteFile(downloadPath string) (io.ReadCloser, e
 		return nil, err
 	}
 	httpClientsDetails := ds.ArtDetails.CreateHttpClientDetails()
-	return ds.client.ReadRemoteFile(readPath, httpClientsDetails, ds.Retries)
+	return ds.client.ReadRemoteFile(readPath, httpClientsDetails)
```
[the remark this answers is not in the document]
barbelity: Removed the unused field.

httpclient/client.go
```diff
@@ -496,40 +538,68 @@ func extractAndMergeChunks(chunksPaths []string, flags ConcurrentDownloadFlags,
...
+			// Perform retry
+			return true, nil
+		},
+	}
+
+	err = retryExecutor.Execute()
+	return
+}
+
+func (jc *HttpClient) downloadFileRange(flags ConcurrentDownloadFlags, start, end int64, currentSplit int, logMsgPrefix string, httpClientsDetails httputils.HttpClientDetails) (string, *http.Response, error) {
```
eyalbe4: It might be better to use named returned values in this method. This should make it easier to understand what the first returned value is.

httpclient/client.go
```diff
@@ -219,23 +232,52 @@ func (jc *HttpClient) ReadRemoteFile(downloadPath string, httpClientsDetails htt
 }
 
 func (jc *HttpClient) DownloadFile(downloadFileDetails *DownloadFileDetails, logMsgPrefix string, httpClientsDetails httputils.HttpClientDetails, retries int, isExplode bool) (*http.Response, error) {
-	resp, _, err := jc.downloadFile(downloadFileDetails, logMsgPrefix, true, httpClientsDetails, retries, isExplode)
+	resp, _, err := jc.downloadFileWithRetries(downloadFileDetails, logMsgPrefix, true, httpClientsDetails, retries, isExplode)
 	return resp, err
 }
 
 func (jc *HttpClient) DownloadFileNoRedirect(downloadPath, localPath, fileName string, httpClientsDetails httputils.HttpClientDetails) (*http.Response, string, error) {
```
eyalbe4: This method does not use retries. I think we should make that all download methods in the client use retries, with a default value of 3, unless configured otherwise.
I think with the uploads we're okay.

httpclient/client.go
```diff
@@ -219,23 +232,52 @@ func (jc *HttpClient) ReadRemoteFile(downloadPath string, httpClientsDetails htt
...
 	downloadFileDetails := &DownloadFileDetails{DownloadPath: downloadPath, LocalPath: localPath, FileName: fileName}
-	return jc.downloadFile(downloadFileDetails, "", false, httpClientsDetails, 0, false)
+	return jc.downloadFile(downloadFileDetails, "", false, httpClientsDetails, false)
 }
 
-func (jc *HttpClient) downloadFile(downloadFileDetails *DownloadFileDetails, logMsgPrefix string, followRedirect bool,
+func (jc *HttpClient) downloadFileWithRetries(downloadFileDetails *DownloadFileDetails, logMsgPrefix string, followRedirect bool,
```
eyalbe4: Since I think all downloads should be with retries, I think we should not add "...WithRetries" to the name of any of the download methods.
The naming convention of the upload methods seems to be good, so maybe we should use the same naming convention of the download methods as well.

utils/retry/retryexecutor.go
```diff
@@ -0,0 +1,58 @@
+package retry
```
eyalbe4: I think we can remove the *retry* package, because it includes only retryexecutor.go and retryexecutor_test.go

utils/retry/retryexecutor.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"fmt"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+	"time"
+)
+
+const retryMessage = "%sAttempt %v - %s"
```
eyalbe4: The convention for constants in go is a leading upper case. Actually, since this template of the message is used only once, maybe we don't need a constant for it.

utils/retry/retryexecutor.go
```diff
@@ -0,0 +1,58 @@
...
+
+type ExecutionHandlerFunc func() (bool, error)
+
+type RetryExecutor struct {
+	// The amount of retries to perform.
+	MaxRetries int
+
+	// Time period between retries.
```
eyalbe4: // Time period between retries. --> // Number of seconds to sleep between retries.

utils/retry/retryexecutor.go
```diff
@@ -0,0 +1,58 @@
...
+		shouldRetry, err = runner.ExecutionHandler()
+
+		// If should not retry, return
+		if !shouldRetry {
+			return err
+		}
+
+		log.Warn(runner.getLogRetryMessage(i, err))
+		// Going to sleep for RetryInterval seconds
+		if i < runner.MaxRetries {
```
eyalbe4: To make sure that we do not call the sleep function with a zero interval value:
if i < runner.MaxRetries { --> if runner.RetriesInterval) > 0 && i < runner.MaxRetries {

eyalbe4:
LIKE @barbelity!
See my inline suggestions.

httpclient/client.go
```diff
@@ -154,15 +149,43 @@ func setRequestHeaders(httpClientsDetails httputils.HttpClientDetails, size int6
...
-		resp, body, err = jc.doUploadFile(localPath, url, httpClientsDetails)
-		if resp != nil && resp.StatusCode < 500 {
-			// No error and status < 500
-			return
-		}
-		log.Warn("Upload attempt #", i, "to", url, "failed -", getFailureReason(resp, err))
+	retryExecutor := retry.RetryExecutor{
+		RetriesNum:      retries,
+		RetriesInterval: 0,
+		RetryMessage:    fmt.Sprintf("Failure occurred while uploading to %s", url),
```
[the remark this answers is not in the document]
barbelity: Changed

httpclient/client.go
barbelity: Changed to named-returned values.

httpclient/client.go
barbelity: Changed DownloadFileNoRedirect to uses the RetryExecutor.

utils/retry/retryexecutor.go
barbelity: Fixed.

utils/retry/retryexecutor.go
barbelity: Done

utils/retry/retryexecutor.go
barbelity: Removed

PR #39, opened 2019-02-18
Fix pr38 comments

barbelity:


artifactory/services/buildinfo.go
```diff
@@ -66,3 +75,28 @@ func (bis *BuildInfoService) GetBuildInfo(params BuildInfoParams) (*buildinfo.Bu
...
+		return nil
+	}
+	httpClientsDetails := bis.GetArtifactoryDetails().CreateHttpClientDetails()
+	utils.SetContentType("application/vnd.org.jfrog.artifactory+json", &httpClientsDetails.Headers)
+	log.Info("Deploying build info...")
+	resp, body, err := bis.client.SendPut(bis.ArtDetails.GetUrl()+"api/build/", content, &httpClientsDetails)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusNoContent {
```
eyalbe4: I think we should change it to:
```if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusNoContent {```
to support a future response code in Artifactory for this request.

eyalbe4:
Looks good @barbelity!
See my inline suggestion.

artifactory/services/buildinfo.go
barbelity: Added support for StatusCreated response code.

PR #40, opened 2019-02-19
initialize upload params with non-nil embedded struct pointer

figadore:
Following the docs, 
```
    params := services.NewUploadParams()
    params.Pattern = "repo/*/*.zip"
``` 
gives `panic: runtime error: invalid memory address or nil pointer dereference`
It seems that the embedded struct pointer needs to be initialized before its properties can be assigned values.

Notes: 
* This is my first PR to this repo. 
* I don't know `go` very well yet.
* I didn't see any tests around this area to update
* I'm not sure whether other areas may have a similar issue

### DIFF
--- a/artifactory/services/upload.go
+++ b/artifactory/services/upload.go
@@ -530,5 +530,8 @@ func (us *UploadService) createFolderInArtifactory(artifact UploadData) error {
 }
 
 func NewUploadParams() UploadParams {
-	return UploadParams{}
+	params := UploadParams{}
+	var commonParams utils.ArtifactoryCommonParams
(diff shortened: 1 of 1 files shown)


barbelity:
Thanks @figadore for this PR.
As this issue was reported by [several users](https://github.com/jfrog/jfrog-client-go/issues/48), and reoccurred in multiple places, we have merged #50 to the [dev branch](https://github.com/jfrog/jfrog-client-go/tree/dev).
Therefore, this PR will not be merged.
We would love to hear your feedback for this fix.

PR #44, opened 2019-03-18
Support multiple UploadParams in upload service

barbelity:


artifactory/manager.go
```diff
@@ -130,12 +130,11 @@ func (sm *ArtifactoryServicesManager) DeleteProps(params services.PropsParams) (
 	return setPropsService.DeleteProps(params)
 }
 
-func (sm *ArtifactoryServicesManager) UploadFiles(params services.UploadParams) (artifactsFileInfo []utils.FileInfo, totalUploaded, totalFailed int, err error) {
+func (sm *ArtifactoryServicesManager) UploadFiles(params ...services.UploadParams) (artifactsFileInfo []utils.FileInfo, totalUploaded, totalFailed int, err error) {
```
eyalbe4: The README should be updated with the new API signature.

artifactory/services/upload.go
```diff
@@ -26,7 +26,6 @@ type UploadService struct {
 	DryRun            bool
 	Threads           int
 	MinChecksumDeploy int64
-	Retries           int
```
eyalbe4: The *Retries* option has been removed from the upload service. 
This causes a bit of inconsistency with the download service, which still includes the *Retries* option.
I think that it make more sense for the *Retries* to be set in the service level than in the upload / download method level, but we can discuss this further.

artifactory/manager.go
barbelity: The README is not affected by this change.

artifactory/services/upload.go
barbelity: This will be fixed in a future development which redesigns the way of working with services and params.

PR #49, opened 2019-03-31
Support retrieval of a list of files with specific extension from specific path

AlexeiVainshtein:


utils/io/fileutils/files.go
```diff
@@ -133,6 +133,32 @@ func ListFilesRecursiveWalkIntoDirSymlink(path string, walkIntoDirSymlink bool)
 	return
 }
 
+// Return the list of files in the specified path that has the extension
+func ListFilesWithSpecificExtension(path, ext string) ([]string, error) {
```
eyalbe4: I suggest shortening the function name to *ListFilesWithExtension*.

utils/io/fileutils/files.go
```diff
@@ -133,6 +133,32 @@ func ListFilesRecursiveWalkIntoDirSymlink(path string, walkIntoDirSymlink bool)
 	return
 }
 
+// Return the list of files in the specified path that has the extension
```
eyalbe4: // Return the list of files in the specified path that has the extension -->
// Return all files with the specified extension in the specified path. Not recursive.

utils/io/fileutils/files.go
```diff
@@ -133,6 +133,32 @@ func ListFilesRecursiveWalkIntoDirSymlink(path string, walkIntoDirSymlink bool)
...
+func ListFilesWithSpecificExtension(path, ext string) ([]string, error) {
+	sep := GetFileSeparator()
```
eyalbe4: I think that the *GetFileSeparator()* function shoukd be removed from thr code base and replaced with ```os.PathSeparator```

utils/io/fileutils/files.go
```diff
@@ -133,6 +133,32 @@ func ListFilesRecursiveWalkIntoDirSymlink(path string, walkIntoDirSymlink bool)
...
+	for _, f := range files {
+		if !strings.HasSuffix(f.Name(), ext) {
+			continue
+		}
+		filePath := path + f.Name()
+		exists, err := IsFileExists(filePath, false)
+		if err != nil {
+			return nil, err
+		}
+		if exists || IsPathSymlink(filePath) {
```
eyalbe4: I belive that you are interested in finding only files or symlinks to files.
I'm not 100% sure, but there's a chance that the above line will also find symlinks to directories. Worth checking just to be sure.

utils/io/fileutils/files.go
AlexeiVainshtein: Done

utils/io/fileutils/files.go
AlexeiVainshtein: Done

utils/io/fileutils/files.go
AlexeiVainshtein: Per our discussion, I'm keeping the function and just replacing with string(os.PathSeparator)

utils/io/fileutils/files.go
AlexeiVainshtein: I've changed it to ignore directories.

PR #51, opened 2019-04-04
feature: Add the ability to list an uploaded file for download

djgilcrease:
feature: Add the ability to list an uploaded file for download https://www.jfrog.com/confluence/display/BT/Bintray+REST+API#BintrayRESTAPI-FileinDownloadList


djgilcrease:
Any feedback?

eyalbe4:
@djgilcrease,
Thank toy for creating PR! Looks like its review has been waiting for a long time. My apologies for this!
Can you please join our contributors community by signing [JFrog's CLA](https://secure.echosign.com/public/hostedForm?formid=5IYKLZ2RXB543N) so that we can merge your code?

djgilcrease:
ping @eyalbe4

djgilcrease:
ping @eyalbe4 any update?

bintray/services/upload.go
```diff
@@ -96,6 +99,15 @@ func (us *UploadService) uploadFiles(uploadDetails *UploadParams, artifacts []cl
 					if uploaded {
 						uploadCount[threadId]++
 					}
+					if uploadDetails.ListDownloads {
+						// This needs to not sleep, but since the publish is async we need to for now
+						time.Sleep(15 * time.Second)
```
eyalbe4: I understand the reason for putting this 15 seconds sleep, but I'm not sure this solution is reliable enough. We're not sure whether 15 seconds is enough or maybe it is too much and causes the operations to be delayed when it could have finished sooner.
Here's a suggested alternative for this:

First, we can send the "add to download list" requests after all files are uploaded, in the main routine, instead of after every file upload as part fo the parallel upload routines. This can increase the chances that each individual file is already published by the time we try to publish it.

The above still does not guarantee that the file is published by the time we try to list it. We can add a retry mechanism as follows:
1. Try to list all published file in a loop (with no sleep).
2. Only if there some files which are unpublished, sleep for a second, and try again #1 (for the remaining files only).

bintray/services/upload.go
```diff
@@ -168,13 +180,29 @@ func uploadFile(artifact clientutils.Artifact, url, logMsgPrefix string, bintray
 	return resp.StatusCode == http.StatusCreated || resp.StatusCode == http.StatusOK, nil
 }
 
+func listFile(url, logMsgPrefix string, bintrayDetails auth.BintrayDetails) (bool, error) {
```
eyalbe4: I suggest renaming this function to *SownInDownloadList*.

bintray/services/upload.go
```diff
@@ -44,13 +46,14 @@ type UploadParams struct {
...
-	Override  bool
-	Publish   bool
-	Deb       string
+	UseRegExp     bool
+	Flat          bool
+	Recursive     bool
+	Explode       bool
+	Override      bool
+	Publish       bool
+	ListDownloads bool
```
eyalbe4: I suggest renaming *ListDownloads* to *ShowInDownloadList*.
Also, let's update the README with this addition.

eyalbe4:
@djgilcrease,
Thanks for changing the implementation.
I see however the following issues with the new design. Please let me know what you think:

1. We try to add files to the download list, even if they failed to be uploaded.
2. There's a new go routine created for every "show in download list" request, which is okay, but not a must in my opinion. 
I guess we could keep this part sequential, to keep the code simpler (I would go with this approach for now). If however we do want to go concurrent here (I'm good with approach too), I think we should respect the threads limit sent to the API and avoid creating X go routines for X files uploaded. 
3. Any reason for the 15 seconds sleep at the end?
4. I suggest not setting a timeout for the "show in download list" request. We don't do this for any other request sent to Bintray.

djgilcrease:
> @djgilcrease,
> Thanks for changing the implementation.
> I see however the following issues with the new design. Please let me know what you think:
> 
> 1. We try to add files to the download list, even if they failed to be uploaded.

Fixed with latest changes

> 2. There's a new go routine created for every "show in download list" request, which is okay, but not a must in my opinion.
>    I guess we could keep this part sequential, to keep the code simpler (I would go with this approach for now). If however we do want to go concurrent here (I'm good with this approach too), I think we should respect the threads limit sent to the API and avoid creating X go routines for X files uploaded.

Fixed with latest changes

> 3. Any reason for the 15 seconds sleep at the end?

Forgot to delete it. Fixed with latest changes

> 4. I suggest not setting a timeout for the "show in download list" request. We don't do this for any other request sent to Bintray.

This is required or you could loop forever since it retries the list for download over any over while waiting for the upload to show up on the server side. I figure if it takes more then 30 seconds for it to show up it will not for some reason, if this is not true we still need some timeout so we can exit to loop

eyalbe4:
@djgilcrease,
Looks like we have only the 4th item to discuss.
I think that the way the timeout mechanism is currently implement won't properly. Here's why - 
Here's a simplified version of the code you wrote:
```
package main

import (
	"context"
	"fmt"
	"time"
)

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
	for {
		select {
		default:
			SownInDownloadList()
			break

		case <-ctx.Done():
			fmt.Println("error")
			break
		}
	}
}

func SownInDownloadList() {
	for i := 0; i < 10; i++ {
		fmt.Println(i)
		time.Sleep(5 * time.Second)
	}
}
```
You can actually copy and run it. Since you're not calling the *SownInDownloadList* from a go routine, it will not timeout after 30 seconds.
Actually, if there's a real need for a client timeout mechanism, it should be implemented in the http client level and not as it is here (and we should already have a timeout for requests sent from the client). Why is the "list downloads" request unique in that sense? 

Also, I'm now more convinced that we should do the listing sequentially, at least as a first stage, and add concurrency if we notice that the listing takes a long time.

djgilcrease:
> @djgilcrease,

> You can actually copy and run it. Since you're not calling the _SownInDownloadList_ from a go routine, it will not timeout after 30 seconds.

of course it wont you have the loop inside `ShowDownloadList` going for 50 seconds. That is NOT why the cancel is there. It is there because `ShowDownloadList` will fail for at least 10 to 15 seconds since the upload is async on the server side, so even though all the bits have been transferred and a response made from the server that it was successfully uploaded the call to list the artifact will error, thus return `listed = false`. If the call to list the artifact errors for 30 seconds in a row it is obviously not going to succeed so we need to break out of the infinite loop.

Even if we did this sequentially we would need this infLoop with a timeout on each artifact unless you want to go back to the blanket `time.Sleep(15*time.Seconds)` that I had before.

Here is a better example https://play.golang.org/p/49PkjDO5Fh2

eyalbe4:
OK @djgilcrease,
I think I understand now what you're trying to achieve. I thought earlier that you were trying to avoid a case where the request gets stuck, but you're just waiting for the files to be published. In that case, I suggest a bit of different solution:
Let's simply loop for up to 30 times -
On each iteration, try to list. If listed successfully, break.
If failed, sleep for 1 second. 
I think that this solution is a bit simpler. I also prefer to do it sequentially, unless we see that doing it contractually provides noticable improvement.

djgilcrease:
ok, addressed the issue and turned listing into a sequential action.

bintray/services/upload.go
```diff
@@ -76,39 +79,69 @@ func (us *UploadService) uploadFiles(uploadDetails *UploadParams, artifacts []cl
...
+				if !us.DryRun {
+					listUrl := us.BintrayDetails.GetApiUrl() + path.Join(
+						"file_metadata",
+						uploadDetails.Subject,
+						uploadDetails.Repo, artifact.TargetPath)
+
+					var listed bool
+					// retry loop, will retry to list uploaded artifacts
+					for j := 0; j < 30; j++ {
+						if listed, _ = SownInDownloadList(listUrl, us.BintrayDetails); listed {
```
eyalbe4: I suggest not to ignore the error returned from the function. In case it is not nil, log it (simply to what we do if the upload fails) and break out of the loop.

bintray/services/upload.go
```diff
@@ -76,39 +79,69 @@ func (us *UploadService) uploadFiles(uploadDetails *UploadParams, artifacts []cl
...
+					uploadedArtifacts[threadId] = append(uploadedArtifacts[threadId], artifacts[j])
 				}
 			}
 			wg.Done()
 		}(i)
 	}
 	wg.Wait()
 
+	if uploadDetails.ShowInDownloadList {
+		// even though we are not running the list for download in go routines we need this outer loop
```
eyalbe4: Great comment. For consistency reasons., let's start with an upper-case (even --> Even).

bintray/services/upload.go
```diff
@@ -76,39 +79,69 @@ func (us *UploadService) uploadFiles(uploadDetails *UploadParams, artifacts []cl
...
+		for i := 0; i < us.Threads; i++ {
+			for _, artifact := range uploadedArtifacts[i] {
+				if !us.DryRun {
+					listUrl := us.BintrayDetails.GetApiUrl() + path.Join(
+						"file_metadata",
+						uploadDetails.Subject,
+						uploadDetails.Repo, artifact.TargetPath)
+
+					var listed bool
+					// retry loop, will retry to list uploaded artifacts
```
eyalbe4: For consistency reasons., let's start with an upper-case (retry --> Retry).

PR #52, opened 2019-04-09
Introduce generic progress bar

RobiNino:
Introduce a new progress bar to upload/download to/from artifactory


utils/log/progress.go
```diff
@@ -0,0 +1,11 @@
+package log
+
+import "io"
+
+type ProgressBar interface {
```
eyalbe4: I suggest renaming *ProgressBar* *Progress*, because people can implement there own progress indicators and not necessarily implement a progress bar.
I also suggest renaming all the *progressBar* variables throughout the code to *progress*.
I'd also suggest adding a description to this interface, explaining that it is used for file transfer progress indication.

utils/log/progress.go
```diff
@@ -0,0 +1,11 @@
...
+	NewBar(total int64, prefix, description string) (barId int)
```
eyalbe4: NewBar --> New

utils/log/progress.go
```diff
@@ -0,0 +1,11 @@
...
+	NewBarReplacement(replaceBarId int, prefix, description string) (barId int)
```
eyalbe4: replaceBarId --> replaceId
NewBarReplacemen --> NewReplacement
barId --> id

The above method and vars probably need a description.

eyalbe4:
Looks awesome @RobiNino!
See my inline suggestions and let me know what you think.
The README should also be updated with the API changes.

eyalbe4:
@RobiNino,
I think that *progress.go* should be moved out of the *log* package, because it is not actually related to logging. Maybe we can put it directly under the *io* package.

PR #56, opened 2019-04-23
Artifactory and Bintray service managers update

barbelity:


eyalbe4:
@barbelity,
I suggest rephrasing the following in README.md from:
```
Several jfrog-client-go operations are storing data on the file-system. By default, the temporary
data will be stored inside 'os.TempDir()'. In order to set a different location for these files, 
simply do the following:
```
to:
```
The default temp dir used is  'os.TempDir()'. Use the following API to set a new temp dir:
```

README.md
```diff
@@ -63,7 +60,10 @@ simply do the following:
     params.Explode = false
     params.Deb = ""
     params.Symlink = false
-    params.Retries = 3
+    // Retries default value: 3
+    params.Retries = 5
+    // MinChecksumDeploy = 10400
```
eyalbe4: // MinChecksumDeploy = 10400 --> // MinChecksumDeploy default value: 10400

PR #57, opened 2019-05-05
Util for extracting username from Artifactory's access-token

barbelity:


artifactory/auth/authutils.go
```diff
@@ -0,0 +1,53 @@
...
+	if len(tokenParts) != 3 {
+		return "", errorutils.CheckError(errors.New("Received invalid access-token."))
+	}
+	payload, _ := base64.RawStdEncoding.DecodeString(tokenParts[1])
+
+	// Unmarshal json.
+	var tokenPayload tokenPayload
+	err := json.Unmarshal(payload, &tokenPayload)
+	if err != nil {
+		fmt.Println(err.Error())
```
eyalbe4: Do we need this?

artifactory/auth/authutils.go
```diff
@@ -0,0 +1,53 @@
...
+
+func ExtractUsernameFromAccessToken(token string) (string, error) {
+	// Separate token parts.
+	tokenParts := strings.Split(token, ".")
+
+	// Decode the payload.
+	if len(tokenParts) != 3 {
+		return "", errorutils.CheckError(errors.New("Received invalid access-token."))
+	}
+	payload, _ := base64.RawStdEncoding.DecodeString(tokenParts[1])
```
eyalbe4: This method returns an error which should be probably checked.

artifactory/auth/authutils.go
```diff
@@ -0,0 +1,53 @@
...
+	if tokenPayload.Subject == "" {
+		return "", errorutils.CheckError(errors.New("Could not extract subject from the provided access-token."))
+	}
+
+	// Extract username from subject.
+	usernameStartIndex := strings.LastIndex(tokenPayload.Subject, "/")
+	if usernameStartIndex < 0 {
+		return "", errorutils.CheckError(errors.New(fmt.Sprintf("Could not extract username from access-token's subject: %s", tokenPayload.Subject)))
+	}
+	username := tokenPayload.Subject[strings.LastIndex(tokenPayload.Subject, "/")+1:]
```
eyalbe4: ```usernameStartIndex``` which has already been calculated can be used here instead of ```strings.LastIndex(tokenPayload.Subject, "/")```

artifactory/auth/authutils.go
barbelity: No, removed.

artifactory/auth/authutils.go
barbelity: Added an error check.

PR #58, opened 2019-05-08
Implement File service

btashton:
File service manager enables get access to the following APIs:
 * File Info
 * File Statistics
 * Item Properties
 * Item Last Modified

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ simply do the following:
 ```
     rtDetails := auth.NewArtifactoryDetails()
     rtDetails.SetUrl("http://localhost:8081/artifactory")
-    rtDetails.SetSshKeysPath("path/to/.ssh/")
+    rtDetails.SetSshKeyPath("path/to/.ssh/")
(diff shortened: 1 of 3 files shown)


eyalbe4:
Thank you for this PR @btashton!
We prefer not to support the FileInfo REST API, but provide this information through AQL. 
The same goes for the *get last modified* and *get statistics* APIs. 
This is actually what the rtManager.SearchFiles API already does - we just need to return additional fields for each file, by enhancing the AQL.
As for reading properties, the *rtManager.SearchFiles(params)* API already returns the properties of files.
Please let me know if you'd like to further discuss the details of the implementation.
Thanks!

btashton:
Fair enough, I'll take a look at the AQL interface and see what if anything needs to be extended to provides the interface I need. I seem to recall running into an issue using AQL for this, but it has been a while.

eyalbe4:
@btashton,
Please let me know if you'd like to keep this PR open.

PR #60, opened 2019-05-26
Change findupstream signature and tests

AlexeiVainshtein:


utils/io/fileutils/files_test.go
```diff
@@ -50,10 +50,14 @@ func TestGetFileOrDirPathFile(t *testing.T) {
 	}
```
eyalbe4: I suggets we add one more test to this function, in which the file is expected not exist:
```
	if exists {
		t.Error("File go.mod is supposed to be missing, but it exists.")
	}
```

AlexeiVainshtein: Done I've added a test also for a file that shouldn't be found.

PR #62, opened 2019-05-27
Support for publishing info file

AlexeiVainshtein:


artifactory/services/go/publishzipandmod.go
```diff
@@ -30,35 +37,39 @@ func (pwa *publishZipAndModApi) PublishPackage(params GoParams, client *rthttpcl
...
 	}
-	err = createUrlPath(moduleId[0], params.GetVersion(), params.GetProps(), ".mod", &url)
+	return nil
+}
+
+func addGoVersion(version string, urlPath *string) {
+	*urlPath += ";go.version=" + url.QueryEscape(version)
+}
+
+func (pwa *publishZipAndModApi) upload(localPath, moduleId, version, props, ext string, urlPath string) error {
```
eyalbe4: The arguments type *string* appears twice, instead of once in the method signature.
I also suggest addign a description for the arguments, specifically for **ext**.

AlexeiVainshtein: Done

PR #63, opened 2019-06-02
Change version api

AlexeiVainshtein:


artifactory/services/go/publishwithmatrixparams.go
```diff
@@ -20,12 +20,11 @@ type publishWithMatrixParams struct {
 func (pwmp *publishWithMatrixParams) isCompatible(artifactoryVersion string) bool {
```
eyalbe4: I suggest adding a test function, which will test this function with a few Artifactory versions.

AlexeiVainshtein: Its being tested in the go_test.go file. There is a test function to check the compatible publisher. The compatible publisher checks the isCompatible of each one that using the interface. I think it's already covered there.

utils/version/version.go
```diff
@@ -52,8 +52,10 @@ func (version *Version) Compare(ver1 string) int {
 	return 0
 }
 
-func (version *Version) AtLeast(artifactoryVersion string) bool {
-	if version.Compare(artifactoryVersion) < 0 && artifactoryVersion != utils.Development {
+// Compares the min version with the version of Artifactory in the version struct
+// Returns false if the min version is bigger then Artifactory version used. Otherwise returns true.
```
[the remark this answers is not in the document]
AlexeiVainshtein: Done

PR #65, opened 2019-06-24
Support wildcard in repositories

yahavi:


eyalbe4:
Awesome work @yahavi!

PR #71, opened 2019-07-26
Fix typo in named return value

nmiyake:


eyalbe4:
Than you for this @nmiyake!
It looks like https://github.com/jfrog/jfrog-client-go/pull/72 already fixes this, in addition to a few other typos.

PR #75, opened 2019-08-07
artifactory : add support for exclude-props in aql build

liron-shalom:


artifactory/services/delete.go
```diff
@@ -122,3 +130,30 @@ func (ds *DeleteParams) SetIncludeDirs(includeDirs bool) {
 func NewDeleteParams() DeleteParams {
 	return DeleteParams{ArtifactoryCommonParams: &utils.ArtifactoryCommonParams{}}
 }
+
+func removeDirsContainArtifactsWithPropsExcluded(specFile utils.ArtifactoryCommonParams, ds *DeleteService, deleteCandidates []utils.ResultItem) ([]utils.ResultItem, error) {
```
eyalbe4: I sugest adding the following description to this function:

*This function receives as an argument the list of files and folders to be deleted from Artifactory.
In case the search params used to create this list included excludeProps, then we might need to remove some directories from this list. These directories are removed from this list because they must not be deleted from Artifactory.
These directories must be deleted, because they include files, which should not be deleted, because of the excludeProps params.*

eyalbe4: I suggest adding a unit test for this function.

artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -111,25 +111,41 @@ func prepareSearchPattern(pattern string, repositoryExists bool) string {
...
+	for _, v := range excludeProperties.Properties {
+		excludePropsQuery += buildExcludedKeyValQueryPart(v.Key, v.Value) + `,`
+	}
+	if len(excludeProperties.Properties) > 1 {
+		excludePropsQuery = `"$or":[` + strings.TrimSuffix(excludePropsQuery, ",") + `],`
+	} else if len(excludeProperties.Properties) == 1 {
+		excludePropsQuery = strings.TrimPrefix(excludePropsQuery, "{")
+		excludePropsQuery = strings.TrimSuffix(excludePropsQuery, "},") + `,`
+	}
+	return propsQuery + excludePropsQuery, nil
```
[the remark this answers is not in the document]
liron-shalom: We took the curly brackets outside the build key-value function.
We'll add all the needed punctuation in the calling function.

Issue #82, opened 2019-09-27
Add support for client certificates


janbernloehr:
In many corporate environments, client certificates are used in addition to classic user credentials (dont blame the messenger).

It would thus be nice to have support in the jfrog clients for client certificates.

PR #83, opened 2019-09-27
Add support for client certificates.

janbernloehr:
This add support for client certificates fixing #82.

### DIFF
--- a/artifactory/auth/rtdetails.go
+++ b/artifactory/auth/rtdetails.go
@@ -26,6 +26,8 @@ type ArtifactoryDetails interface {
 	GetPassword() string
 	GetApiKey() string
 	GetAccessToken() string
+	GetClientCertificatePath() string
+	GetClientCertificateKeyPath() string
 	GetSshUrl() string
 	GetSshKeyPath() string
 	GetSshPassphrase() string
@@ -37,6 +39,8 @@ type ArtifactoryDetails interface {
 	SetPassword(password string)
 	SetApiKey(apiKey string)
 	SetAccessToken(accessToken string)
+	SetClientCertificatePath(certificatePath string)
+	SetClientCertificateKeyPath(certificatePath string)
 	SetSshUrl(url string)
 	SetSshKeyPath(sshKeyPath string)
 	SetSshPassphrase(sshPassphrase string)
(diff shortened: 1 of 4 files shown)


artifactory/auth/rtdetails.go
```diff
@@ -218,7 +241,10 @@ func (rt *artifactoryDetails) GetVersion() (string, error) {
 }
 
 func (rt *artifactoryDetails) getArtifactoryVersion() (string, error) {
-	client, err := httpclient.ClientBuilder().Build()
+	client, err := httpclient.ClientBuilder().
+		SetClientCertificatePath((*rt).GetClientCertificatePath()).
+		SetClientCertificateKeyPath((*rt).GetClientCertificateKeyPath()).
+		Build()
```
eyalbe4: The above can be simplified to:
```
	client, err := httpclient.ClientBuilder().
		SetClientCertificatePath(rt.GetClientCertificatePath()).
		SetClientCertificateKeyPath(rt.GetClientCertificateKeyPath()).
		Build()
```

httpclient/clientBuilder.go
```diff
@@ -33,13 +45,27 @@ func (builder *httpClientBuilder) Build() (*HttpClient, error) {
...
 	if err != nil {
 		return nil, errorutils.CheckError(errors.New("Failed creating HttpClient: " + err.Error()))
 	}
+	if builder.clientCertificatePath != "" {
+		cert, err := tls.LoadX509KeyPair(builder.clientCertificatePath, builder.clientCertificateKeyPath)
+		if err != nil {
+			return nil, errorutils.CheckError(errors.New("Failed creating HttpClient: " + err.Error()))
+		}
+		transport.TLSClientConfig.Certificates = []tls.Certificate{cert}
+	}
```
eyalbe4: The above code is duplicated inside this method.
Let's take the repeating code into a different method and reuse it.

janbernloehr:
I did as you asked and renamed all occurances of ClientCertificate to ClientCert. This, however, is now inconsistent with several other functions, e.g. `SetCertificatesPath`, and also go's http library always uses `Certificate` instead of `Cert`.

We can keep the sort version with cert if you like it more, just wanted to point this out.

PR #93, opened 2019-11-05
 Fix .. (the parent directory) in upload pattern.

Postyy:


utils/utils.go
```diff
@@ -130,11 +131,22 @@ func PrepareLocalPathForUpload(localPath string, useRegExp bool) string {
 		localPath = localPath[3:]
 	}
 	if !useRegExp {
-		localPath = pathToRegExp(localPath)
+		localPath = pathToRegExp(cleanPath(localPath))
 	}
 	return localPath
 }
 
+// Clean /../ | /./ and return the double backslash in case of windows path
```
eyalbe4: I suggest changing this comment to the following:
 // Clean /../ | /./ using filepath.Clean.

utils/utils.go
```diff
@@ -130,11 +131,22 @@ func PrepareLocalPathForUpload(localPath string, useRegExp bool) string {
...
+func cleanPath(path string) string {
+	temp := path[len(path)-1:]
+	path = filepath.Clean(path)
+	if temp == `\` || temp == "/" {
+		path += temp
+	}
+	path = strings.Replace(path, `\`, `\\`, -1)
```
eyalbe4: Let's add the following comment to describe this line:
// Since filepath.Clean replaces \\ with \, we revert this action.

utils/utils_test.go
```diff
@@ -63,3 +63,84 @@ func assertSplitWithEscape(str string, expected []string, t *testing.T) {
...
 	}
 }
+
+func TestCleanPath(t *testing.T) {
+	if IsWindows() {
+		parameter := "\\\\foo\\\\baz\\\\..\\\\bar\\\\*"
+		got := cleanPath(parameter)
+		want := "\\\\foo\\\\bar\\\\*"
+		if got != want {
+			t.Errorf("PrepareLocalPathForUpload(%s) == %s, want %s", parameter, got, want)
```
eyalbe4: Shouldn't *PrepareLocalPathForUpload* be replaced with *cleanPath*?
(In multiple occurrences in this function).

utils/utils.go
Postyy: Fixed

PR #101, opened 2019-12-02
Added caddyshack needed features to support latest JFrog CLI

elioengcomp:
- Add GET request with retry on timeout to http client
- Add HEAD request with retry on timeout to http client
- Add retry on timeout to Artifactory version retrieval method
- Add ignore forbidden error flag to Go module publish service


elioengcomp:
See #121 instead.

PR #104, opened 2019-12-09
Add type to dependency build info.

Or-Gabay:


Or-Gabay:
Relate to https://github.com/jfrog/gocmd/pull/23/files

PR #122, opened 2020-02-12
Add support for release bundles in all AQL based commands

yahavi:
Tests: 
https://ci.appveyor.com/project/yahavi/jfrog-client-go


artifactory/services/utils/artifactoryutils.go
```diff
@@ -178,6 +178,42 @@ func parseBuildNameAndNumber(buildIdentifier string) (buildName string, buildNum
 	return buildName, buildNumber
 }
 
+func getBundleNameAndNumberFromBundleIdentifier(bundleIdentifier string) (string, string, error) {
```
eyalbe4: I suggest renaming this function to "parseBundleNameAndVersion", to be compatible with the exiting "parseBuildNameAndNumber" function.
The two function also have some code can be reused. For example. the following loop, but I think there's more:
```
	for i := len(bundle) - 1; i >= 1; i-- {
		bundleVersionsArray = append([]string{bundle[i]}, bundleVersionsArray...)
		if !strings.HasSuffix(bundle[i-1], EscapeChar) {
			bundleName = strings.Join(bundle[:i], Delimiter)
			bundleVersion = strings.Join(bundleVersionsArray, Delimiter)
			break
		}
	}
```

artifactory/services/utils/searchutil.go
```diff
@@ -74,13 +87,23 @@ func SearchBySpecWithAql(specFile *ArtifactoryCommonParams, flags CommonConf, re
...
+			resultsArtifactsIncludeProperties := requiredArtifactProps != NONE && includePropertiesInAqlForSpec(specFile)
+			results, err = filterAqlSearchResultsByBuild(specFile, results, flags, resultsArtifactsIncludeProperties)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		// Filter results by bundle.
+		if specFile.Bundle != "" {
+			results, err = filterAqlSearchResultsByBundle(specFile, results, flags)
```
eyalbe4: It looks like we're using a two AQLs approach when the spec includes a bundle. Do we have to do this? I know that we had to tale this approach for "builds" due to performance reasons, but for bundles, can we have one AQL to include the bundles?

yahavi: This is a difficult question. Maybe @galusben will know the answer?

Issue #127, opened 2020-03-12
Create Token without expiry (ExpiresIn=0) not possible (token is issued with default expiry)


markuslackner:
Currently its not possible to create a token without expiry over the api. The reason is this code:

https://github.com/jfrog/jfrog-client-go/blob/74b10e4efa5cfc819f061cbf3bbba99d6f7cb6b6/artifactory/services/security.go#L139

=> If you specify "0" the expires_in is not send to the server and the default is used.

eyalbe4:
Thanks for reporting this @markuslackner.
Feel free to create a pull request to change this behaviour.

PR #128, opened 2020-03-12
fixed expires_in=0 (never expires)

markuslackner:
see  #127 

Solution slightly changes behaviour for defaults. Consider following:

```
securityService := NewSecurityService(client)
securityService.CreateToken(CreateTokenParams{Username: "test"})
```
Without this MR no `expires_in` is send to the server, with this MR expires_in=0 is sent and the issued token never expires. 

If the API is used with `NewCreateTokenParams` function there is no difference.

```
securityService := NewSecurityService(client)
params := NewCreateTokenParams()
params.Username = "test"
securityService.CreateToken(params)
```

Also negative values are no more sent to the server and so the server default is used.

Maybe thats not the best solution. I am open for ideas!

### DIFF
--- a/artifactory/services/security_test.go
+++ b/artifactory/services/security_test.go
@@ -2,6 +2,7 @@ package services
 
 import "testing"
 
+// Test mapping ExpiresIn to expires_in request value and default handling
 func TestBuildCreateTokenUrlValuesExpiresIn(t *testing.T) {
 	tests := []struct {
 		testName string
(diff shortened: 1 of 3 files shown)


artifactory/services/security_test.go
```diff
@@ -0,0 +1,30 @@
+package services
+
+import "testing"
+
+func TestBuildCreateTokenUrlValuesExpiresIn(t *testing.T) {
```
eyalbe4: I suggest merging this test with the other tokens tests in *artifactorysecurity_test.go*.
Let's also add a description for this test, explaining what it being tested.

markuslackner: Currently the tests are testing the unexported method `buildCreateTokenUrlValues` in [security.go](url). I think migrating them to [artifactorysecurity_test.go](https://github.com/jfrog/jfrog-client-go/blob/dev/tests/artifactorysecurity_test.go) would mean to build integration tests, which need a local artifactory instance.

PR #129, opened 2020-03-17
Support refreshable tokens process

RobiNino:


auth/dsdetails.go
```diff
@@ -183,6 +197,10 @@ func (ds *CommonConfigFields) HandleTokenExpiry(statusCode int, httpClientDetail
 	if statusCode == http.StatusUnauthorized && ds.IsSshAuthentication() {
 		return ds.handleSshTokenExpiry(httpClientDetails)
 	}
+
+	if statusCode == http.StatusUnauthorized && ds.GetAccessToken() != "" && ds.GetTokenRefreshHandler() != nil {
+		return ds.handleAccessTokenExpiry(httpClientDetails)
+	}
```
eyalbe4: The above two if statements can be merged and simplified as foolows:
```
	if statusCode == http.StatusUnauthorized {
		// If an unauthorized ssh connection -> ssh token has expired.
		if ds.IsSshAuthentication() {
			return ds.handleSshTokenExpiry(httpClientDetails)
		}
		if ds.GetAccessToken() != "" && ds.GetTokenRefreshHandler() != nil {
			return ds.handleAccessTokenExpiry(httpClientDetails)
		}
	}
```

PR #133, opened 2020-03-17
Repo services - add more tests + minor fixes

liron-shalom:


tests/utils.go
```diff
@@ -419,7 +420,7 @@ func validateRepoConfig(t *testing.T, repoKey string, params interface{}) bool {
 		return false
 	}
 	for key, value := range paramsMap {
-		if confMap[key] != value {
+		if !reflect.DeepEqual(confMap[key], value) {
 			t.Error(fmt.Sprintf("Key %s: Expected: %v, Actual: %v", key, value, confMap[key]))
 			return false
 		}
```
yahavi: Reflection is forbidden, unless we have no choice.
I suggest remove the bool return value from this function and replace all `if -> error` with assertions.
For example: 
```go
for key, value := range paramsMap {
    assert.NotEqual(t, confMap[key], value)
}
```

tests/artifactorylocalrepository_test.go
```diff
@@ -46,9 +47,126 @@ func localMavenTest(t *testing.T) {
 	if err != nil {
 		t.Error("Failed to update " + repoKey)
 	}
```
yahavi: I suggest replacing all if err != nil to assert.NoError:
```suggestion
      assert.NoError(t, err, "Failed to update " + repoKey)
```

PR #140, opened 2020-03-30
Do not modify params struct while building AQL query.

liron-shalom:


artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -284,10 +284,13 @@ func stringIsInSlice(string string, strings []string) bool {
 }
 
 func prepareFieldsForQuery(fields []string) []string {
-	for i, val := range fields {
-		fields[i] = `"` + val + `"`
+	// Since a slice is basically a pointer, we don't ant to modify the underlying fields array cause it might be used again (like in delete service)
```
eyalbe4: ...don't ant to... --> ...don't want to...

PR #132, opened 2020-03-17
Rename auth/dsdetails to auth/commondetails

yahavi:

PR #143, opened 2020-04-01
Add ssh authentication to service manager

RobiNino:
Also:
Make GetVersion functions use a service.
Rename Wrong file #132

### DIFF
--- a/artifactory/auth/rtdetails.go
+++ b/artifactory/auth/rtdetails.go
@@ -1,23 +1,15 @@
 package auth
 
 import (
-	"encoding/json"
-	"errors"
-	rthttpclient "github.com/jfrog/jfrog-client-go/artifactory/httpclient"
+	"github.com/jfrog/jfrog-client-go/artifactory"
 	"github.com/jfrog/jfrog-client-go/auth"
-	"github.com/jfrog/jfrog-client-go/utils"
-	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+	"github.com/jfrog/jfrog-client-go/config"
 	"github.com/jfrog/jfrog-client-go/utils/log"
-	"net/http"
-	"strings"
-	"sync"
 )
 
 func NewArtifactoryDetails() auth.CommonDetails {
 	return &artifactoryDetails{}
 }
-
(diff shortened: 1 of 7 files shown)


artifactory/auth/rtdetails.go
```diff
@@ -37,29 +29,14 @@ func (rt *artifactoryDetails) GetVersion() (string, error) {
 
 func (rt *artifactoryDetails) getArtifactoryVersion() (string, error) {
 	cd := auth.CommonDetails(rt)
-	client, err := rthttpclient.ArtifactoryClientBuilder().
-		SetCommonDetails(&cd).
+	serviceConfig, err := config.NewConfigBuilder().
+		SetArtDetails(cd).
+		SetCertificatesPath(cd.GetClientCertPath()).
+		SetThreads(0).
```
eyalbe4, line 35: I think that *SetThreads(0)* here is redundant.

auth/commondetails.go
```diff
@@ -233,11 +247,11 @@ func SshTokenRefreshPreRequestInterceptor(fields *CommonConfigFields, httpClient
 	return nil
 }
```
eyalbe4, line 249: In the *SshTokenRefreshPreRequestInterceptor* function in this class, the following comment should be probably changed:
// Obtain a new token and return true (false for error). --> // Obtain a new token.
(because a boolean is not returned there).

distribution/auth/dsdetails.go
```diff
@@ -38,29 +29,14 @@ func (ds *distributionDetails) GetVersion() (string, error) {
 
 func (ds *distributionDetails) getDistributionVersion() (string, error) {
 	cd := auth.CommonDetails(ds)
-	client, err := rthttpclient.ArtifactoryClientBuilder().
-		SetCommonDetails(&cd).
+	serviceConfig, err := config.NewConfigBuilder().
+		SetArtDetails(cd).
+		SetCertificatesPath(cd.GetClientCertPath()).
+		SetThreads(0).
```
eyalbe4, line 35: SetThreads(0) is probably redundant here.

PR #145, opened 2020-04-13
Rename "SetArtDetails" and "CommonDetails" to "SetServiceDetails" and…

eyalbe4:
Following the introduction of the new *VersionService* in https://github.com/jfrog/jfrog-client-go/pull/143, we now have new service - Distribution which also exposes a VersionService.
This change led to an API that looked as follows:
```
func (ds *distributionDetails) getDistributionVersion() (string, error) {
	cd := auth.CommonDetails(ds)
	serviceConfig, err := config.NewConfigBuilder().
		SetArtDetails(cd).
		SetCertificatesPath(cd.GetClientCertPath()).
		SetThreads(0).
		Build()
	sm, err := distribution.New(&cd, serviceConfig)
	if err != nil {
		return "", err
	}
	return sm.GetDistributionVersion()
```
Notice the *SetArtDetails(cd)* API above. The naming is wrong, because we're setting here distribution service, not the artifactory service. Since SetArtDetails is defined by a common interface for Artifactory and Distribution, it makes sense use a more generic name - *SetServiceDetails*.
To keep the naming clear and consistent, *CommonDetails* becomes *ServiceDetails*.

Following the change, the above API becomes:
```
func (ds *distributionDetails) getDistributionVersion() (string, error) {
	cd := auth.ServiceDetails(ds)
	serviceConfig, err := config.NewConfigBuilder().
		SetServiceDetails(cd).
		SetCertificatesPath(cd.GetClientCertPath()).
		Build()
	sm, err := distribution.New(&cd, serviceConfig)
	if err != nil {
		return "", err
	}
	return sm.GetDistributionVersion()
}
```


distribution/manager.go
```diff
@@ -12,16 +12,16 @@ type DistributionServicesManager struct {
 	config config.Config
 }
 
-func New(commonDetails *auth.CommonDetails, config config.Config) (*DistributionServicesManager, error) {
-	err := (*commonDetails).InitSsh()
+func New(ServiceDetails *auth.ServiceDetails, config config.Config) (*DistributionServicesManager, error) {
```
RobiNino, line 15: I suggest changing the function argument to start with a lower case letter, to follow our conventions.

tests/utils.go
```diff
@@ -247,34 +247,34 @@ func failOnHttpClientCreation(err error) {
...
+func GetDistDetails() auth.ServiceDetails {
 	distDetails := distributionAuth.NewDistributionDetails()
 	distDetails.SetUrl(clientutils.AddTrailingSlashIfNeeded(*DistUrl))
 	setAuthenticationDetail(distDetails)
 	return distDetails
 }
 
-func setAuthenticationDetail(commonDetails auth.CommonDetails) {
-	if !fileutils.IsSshUrl(commonDetails.GetUrl()) {
+func setAuthenticationDetail(ServiceDetails auth.ServiceDetails) {
```
RobiNino, line 264: I suggest changing the function argument to start with a lower case letter, to follow our conventions.

eyalbe4: Nice catch. Thanks!

PR #149, opened 2020-04-19
Upload - add VCS properties to a single file only when needed.

liron-shalom:


eyalbe4:
Fixes https://github.com/jfrog/jfrog-cli/issues/618

PR #150, opened 2020-04-19
Service Manager - Add system service.

liron-shalom:
* Change versionService to a method under this service
    * Add a method to get ArtifactoryServiceId


artifactory/manager.go
```diff
@@ -272,10 +272,10 @@ func (sm *ArtifactoryServicesManager) GetReplication(repoKey string) ([]utils.Re
 	return getPushReplicationService.GetReplication(repoKey)
 }
 
-func (sm *ArtifactoryServicesManager) GetArtifactoryVersion() (string, error) {
-	versionService := services.NewVersionService(sm.client)
-	versionService.ArtDetails = sm.config.GetServiceDetails()
-	return versionService.GetArtifactoryVersion()
+func (sm *ArtifactoryServicesManager) CreateSystemService() *services.SystemService {
```
eyalbe4, line 275: It looks like this function as asymmetric compared to the other functions in this file.
Notice that non of the other functions return a service. This is the only function which returns a service. I suggest replacing it with two functions - GetVersion and GetServiceId (both functions will use the new SystemService of course).

artifactory/services/system.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (ss *SystemService) GetJfrogHttpClient() (*rthttpclient.ArtifactoryHttpClient, error) {
+	return ss.client, nil
+}
+
+func (ss *SystemService) IsDryRun() bool {
+	return false
+}
+
+func (ss *SystemService) GetArtifactoryVersion() (string, error) {
```
eyalbe4, line 39: I suggest renaming this function to *GetVersion()*.

artifactory/services/system.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	var version artifactoryVersion
+	err = json.Unmarshal(body, &version)
+	if err != nil {
+		return "", errorutils.CheckError(err)
+	}
+	return strings.TrimSpace(version.Version), nil
+}
+
+func (ss *SystemService) GetArtifactoryServiceId() (string, error) {
```
eyalbe4, line 57: I suggest renaming this function to *GetServiceId()*.

PR #162, opened 2020-05-25
Artifactory Services - Add new service to download files with results reader 

liron-shalom:
* change current service to return only statistics and error
* add new service to download with results reader as an output.


README.md
```diff
@@ -102,6 +102,8 @@ rtManager.UploadFiles(params)
 
 #### Downloading Files from Artifactory
 
+##### Downloading Files without Resutls Reader:
+Using `DownloadFiles()` service, we can download files and get the general statistics of the action (The actual number of files downloaded, and the number of files we expected to download), and the error value if it occurred.
```
eyalbe4, line 106: Using `DownloadFiles()` service... --> Using the `DownloadFiles()` function...

README.md
```diff
@@ -102,6 +102,8 @@ rtManager.UploadFiles(params)
 
 #### Downloading Files from Artifactory
 
+##### Downloading Files without Resutls Reader:
```
eyalbe4, line 105: ##### Downloading Files without Resutls Reader: --> ##### Downloading Files:

README.md
```diff
@@ -119,7 +121,40 @@ params.SplitCount = 2
 // MinSplitSize default value: 5120
 params.MinSplitSize = 7168
 
-rtManager.DownloadFiles(params)
+totalDownloaded, totalExpected, err := rtManager.DownloadFiles(params)
+```
+
+##### Downloading Files with Results Reader:
+Using `DownloadFilesWithResultReader()` service, we can download files and get the general statistics of the action (The actual number of files downloaded, and the number of files we expected to download), and the error value if it occurred.
+In addition, this service provides us with a reader through which we can iterate the downloaded files information
```
eyalbe4, line 129: I suggest changing the description to:
Similar to <link to download files section>, but returns a reader, which allows iterating over the details of the downloaded files. Only files which were successfully downloaded are available by the reader.

README.md
```diff
@@ -119,7 +121,40 @@ params.SplitCount = 2
...
+params.Flat = false
+params.Explode = false
+params.Symlink = true
+params.ValidateSymlink = false
+// Retries default value: 3
+params.Retries = 5
+// SplitCount default value: 3
+params.SplitCount = 2
+// MinSplitSize default value: 5120
+params.MinSplitSize = 7168
```
eyalbe4, line 145: No need to provide the above params, because we're referencing the user to the previous section.

README.md
```diff
@@ -119,7 +121,40 @@ params.SplitCount = 2
...
+reader, totalDownloaded, totalExpected, err := rtManager.DownloadFilesWithResultReader(params)
+```
+Use `reader.NextRecord()` and `FileInfo` type from `utils` package to iterate over the download results.
+Use `reader.Close()` method (preferably using `defer`), to remove the results reader after it is used:
+``` 
+defer reader.Close()
+var file utils.FileInfo
+for e := resultReader.NextRecord(&file); e == nil; e = resultReader.NextRecord(&file) {
+    // Do somthing with file
+    ...
```
eyalbe4, line 156: Let's use ```fmt.Println``` to print the fields of the FileInfo struct.

PR #167, opened 2020-06-08
Support placeholders use in repository name for download, move and copy

barbelity:


utils/utils.go
```diff
@@ -185,24 +185,35 @@ func pathToRegExp(localPath string) string {
...
 //      returns "hello"
 // Example 2:
 //      pattern = "repoA/1(.*)234" ; path = "repoB/1hello234" ; target = "{1}" ; ignoreRepo = true
 //      returns "hello"
 func BuildTargetPath(pattern, path, target string, ignoreRepo bool) (string, error) {
-	if ignoreRepo {
+	asteriskIndex := strings.Index(pattern, "*")
+	slashIndex := strings.Index(pattern, "/")
+	if shouldRemoveRepo(ignoreRepo, asteriskIndex, slashIndex) {
+		// Required when working with virtual repositories as the pattern may contain the virtual-repository name,
```
eyalbe4, line 202: Required when ... --> Required the repository part of the path, when ...

utils/utils.go
```diff
@@ -185,24 +185,35 @@ func pathToRegExp(localPath string) string {
...
+		// but the path will contain the local-repository name.
 		pattern = removeRepoFromPath(pattern)
 		path = removeRepoFromPath(path)
 	}
 	pattern = addEscapingParentheses(pattern, target)
 	pattern = pathToRegExp(pattern)
+	if slashIndex < 0 {
+		// '/' doesn't exists, add an optional trailing-slash to support cases in which the provided pattern
```
eyalbe4, line 210: '/' doesn't exists... --> If '/' doesn't existt...

utils/utils.go
barbelity: What do you think of the following:
`Removing the repository part of the path is required when working with virtual repositories, as the pattern may contain the virtual-repository name, but the path contains the local-repository name.`

utils/utils.go
barbelity: Fixed.

utils/utils.go
eyalbe4: LGTM.

PR #171, opened 2020-06-15
Add exclusions to CopyDir

RobiNino:


utils/io/fileutils/files.go
```diff
@@ -404,6 +410,15 @@ func CopyDir(fromPath, toPath string, includeDirs bool) error {
 	return err
 }
 
+func IsStringInSlice(string string, strings []string) bool {
```
eyalbe4, line 413: I suspsect we can get a nil pointer here.
The *strings* argument can be nil, because we do sent nil to the CopyDir function here - ```err := CopyDir(v, toPath, true, nil)```

RobiNino: range on nil array in go behaves the same as range on empty array

eyalbe4:
> Range on nil slice in go behaves the same as range on empty slice, so there is no risk of panic here. 

OK.

PR #172, opened 2020-06-15
Add log to bp dry-run

RobiNino:


artifactory/services/buildinfo.go
```diff
@@ -82,6 +82,7 @@ func (bis *BuildInfoService) PublishBuildInfo(build *buildinfo.BuildInfo) error
 		return err
 	}
 	if bis.IsDryRun() {
+		log.Info("[Dry run] Outputting Build info preview...")
```
eyalbe4, line 85: Outputting --> Logging

PR #174, opened 2020-06-16
Read vcs revision from packed-refs

RobiNino:
Following [704](https://github.com/jfrog/jfrog-cli/issues/704) in JFrog CLI.


utils/git.go
```diff
@@ -143,11 +159,44 @@ func (m *manager) readRevision() {
...
+			return
+		}
+		defer file.Close()
+
+		scanner := bufio.NewScanner(file)
+		for scanner.Scan() {
+			line := scanner.Text()
+			if strings.HasSuffix(line, ref) {
+				split := strings.Split(line, " ")
+				m.revision = split[0]
```
eyalbe4, line 190: Question - shouldn't we add some validation here? Since I'm not sure what the expected line structure is, I'm not sure abput it. Do we want to raise an error here in case the split returned only one section?
I think it is good to add a comment with the expected line structure.

RobiNino: I added a validation to the split which I think is sufficient, and an error raise. I cannot add a validation on the revision's length since it could be abbreviated. What do you think?

PR #185, opened 2020-08-06
Replication - add comments.

Or-Gabay:


README.md
```diff
@@ -535,6 +535,7 @@ Example of creating repository replication:
 
 ```go
 params := services.NewCreateReplicationParams()
+// Source repository which configuration will be load to.
```
eyalbe4, line 538: Incorrect comment grammar. I suggest changing it to:
// Source replication repository.

The above comment also applies to the README of course.

PR #197, opened 2020-09-06
Artifactory services manager interface

Or-Gabay:


artifactory/emptymanager.go
```diff
@@ -0,0 +1,249 @@
...
+	GetReplication(repoKey string) ([]utils.ReplicationParams, error)
+	GetVersion() (string, error)
+	GetServiceId() (string, error)
+	PromoteDocker(params services.DockerPromoteParams) error
+	Client() *rthttpclient.ArtifactoryHttpClient
+}
+
+type EmptyArtifactoryServicesManager struct {
+}
+
```
eyalbe4, line 66: I'm not sure if go allows this to be complied, but in case it does, you can add the following function:
```
func (esm *EmptyArtifactoryServicesManager) fail() {
	panic("Failed: Method is not implemented")
}
```
and call it from all functions in this struct. For example, the following functions:
```
func (esm *EmptyArtifactoryServicesManager) CreateLocalRepository() *services.LocalRepositoryService {
	panic("Failed: Method is not implemented")
}
```
will be replaced with:
```
func (esm *EmptyArtifactoryServicesManager) CreateLocalRepository() *services.LocalRepositoryService {
	fail()
}
```

artifactory/emptymanager.go
```diff
@@ -0,0 +1,249 @@
...
+	UpdateReplication(params services.UpdateReplicationParams) error
+	DeleteReplication(repoKey string) error
+	GetReplication(repoKey string) ([]utils.ReplicationParams, error)
+	GetVersion() (string, error)
+	GetServiceId() (string, error)
+	PromoteDocker(params services.DockerPromoteParams) error
+	Client() *rthttpclient.ArtifactoryHttpClient
+}
+
+type EmptyArtifactoryServicesManager struct {
```
eyalbe4, line 64: Let's add the following description for this struct:
// By using this struct, you have the option of overriding only some of the ArtifactoryServicesManager interface's methods, but still implement this interface. This comes in very handy for tests.

artifactory/manager.go
```diff
@@ -14,22 +14,22 @@ import (
...
 
-type ArtifactoryServicesManager struct {
+type ArtifactoryServicesManagerImp struct {
 	client   *rthttpclient.ArtifactoryHttpClient
 	config   config.Config
 	progress ioutils.Progress
 }
 
-func New(artDetails *auth.ServiceDetails, config config.Config) (*ArtifactoryServicesManager, error) {
+func New(artDetails *auth.ServiceDetails, config config.Config) (*ArtifactoryServicesManagerImp, error) {
```
eyalbe4, line 23: Can this function return ```ArtifactoryServicesManager``` instead of ```* ArtifactoryServicesManagerImp```?
The same goes for ```NewWithProgress```.

eyalbe4 (changes requested):

artifactory/emptymanager.go
Or-Gabay: Golang does allow this but then I will have to supply a return statement for all methods.

PR #204, opened 2020-09-22
Add support for escaping of commas in upload command

talarian1:
Fixes https://github.com/jfrog/jfrog-cli/issues/786

### DIFF
--- a/artifactory/services/utils/properties.go
+++ b/artifactory/services/utils/properties.go
@@ -42,8 +42,15 @@ func ParseProperties(propStr string, option PropertyParseOptions) (*Properties, 
 
 		switch option {
 		case SplitCommas:
-			for _, val := range strings.Split(values, ",") {
-				props.Properties = append(props.Properties, Property{key, val})
+			splitedValues := strings.Split(values, ",")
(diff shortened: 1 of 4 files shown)


artifactory/services/utils/properties.go
```diff
@@ -42,8 +42,15 @@ func ParseProperties(propStr string, option PropertyParseOptions) (*Properties,
 
 		switch option {
 		case SplitCommas:
-			for _, val := range strings.Split(values, ",") {
-				props.Properties = append(props.Properties, Property{key, val})
+			splitedValues := strings.Split(values, ",")
+			for i, val := range splitedValues {
+				// If \ found it means the original string contains the \, which indicate this "," is part of the value
```
eyalbe4, line 47: // If \ found it means the ... --> // If \ is found, then it means that the ...

eyalbe4 (changes requested):
Let's add the new **BuildProps** property to the README.
See the *Uploading Files to Artifactory* section.

eyalbe4 (changes requested):
See my above comments.

PR #207, opened 2020-09-29
Fix symlink upload target file name

barbelity:


eyalbe4 (changes requested):
For future reference, let'a add to this PR a description, detailing the issue this PR solves.

PR #213, opened 2020-10-15
Added detailed summary support for upload command

talarian1:
Related: https://github.com/jfrog/jfrog-cli-core/pull/23.

### DIFF
--- a/artifactory/services/upload.go
+++ b/artifactory/services/upload.go
@@ -17,6 +17,7 @@ import (
 	clientutils "github.com/jfrog/jfrog-client-go/utils"
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 	ioutils "github.com/jfrog/jfrog-client-go/utils/io"
+	"github.com/jfrog/jfrog-client-go/utils/io/content"
 	"github.com/jfrog/jfrog-client-go/utils/io/fileutils"
 	"github.com/jfrog/jfrog-client-go/utils/io/fileutils/checksum"
 	"github.com/jfrog/jfrog-client-go/utils/io/httputils"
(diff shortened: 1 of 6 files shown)


talarian1:

artifactory/services/upload.go
```diff
@@ -496,7 +502,7 @@ type UploadData struct {
 
 type artifactContext func(UploadData) parallel.TaskFunc
 
-func (us *UploadService) createArtifactHandlerFunc(uploadResult *utils.UploadResult, uploadParams UploadParams) artifactContext {
+func (us *UploadService) createArtifactHandlerFunc(uploadResult *utils.Result, uploadParams UploadParams) artifactContext {
```
eyalbe4, line 505: It looks like the following local variables used in the function, do nor need to be declared implicitly. They can be created when returned from the functions returning them:
```
			var uploaded bool
			var target string
			var artifactFileInfo utils.FileInfo
```

artifactory/services/upload.go
```diff
@@ -90,7 +90,13 @@ func (us *UploadService) performUploadTasks(consumer parallel.Runner, uploadSumm
 	if totalFailed > 0 {
 		log.Error("Failed uploading", strconv.Itoa(totalFailed), "artifacts.")
 	}
-	artifactsFileInfo = utils.FlattenFileInfoArray(uploadSummary.FileInfo)
+	if us.ResultWriter != nil {
+		err = us.ResultWriter.Close()
```
eyalbe4, line 94: It makes more sense not to close the ```ResultWriter``` here. The function that creates it (```UploadFilesWithResultReader```) should be the one that closes with defer.
The same should be done in the download flow, in the ```performTasks``` method, in download.go.
This approach is not always possible, but I think it is good practice to stick to this approach when possible (if you opened it? You'll close it :)).

eyalbe4 (changes requested):

PR #216, opened 2020-10-20
Enable set/delete properties request to be non recursive

talarian1:
Fixes https://github.com/jfrog/jfrog-cli/issues/669

### DIFF
--- a/artifactory/services/utils/searchutil.go
+++ b/artifactory/services/utils/searchutil.go
@@ -94,12 +94,7 @@ func SearchBySpecWithAql(specFile *ArtifactoryCommonParams, flags CommonConf, re
 	}
 	if filteredReader != nil {
 		defer reader.Close()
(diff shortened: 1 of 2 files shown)


artifactory/services/utils/searchutil.go
```diff
@@ -94,12 +94,7 @@ func SearchBySpecWithAql(specFile *ArtifactoryCommonParams, flags CommonConf, re
 	}
 	if filteredReader != nil {
 		defer reader.Close()
-		fetchedProps, err = fetchProps(specFile, flags, requiredArtifactProps, filteredReader)
-		if fetchedProps != nil {
-			defer filteredReader.Close()
-			return fetchedProps, err
-		}
-		return filteredReader, err
+		reader = filteredReader
```
eyalbe4, line 97: Since the logic that decides which reader should be closes is not trivial, let's try to explain it with comments.

eyalbe4 (changes requested):

PR #218, opened 2020-10-25
Add option to pass compareFunction to ListFilesWithExtension

RobiNino:


utils/io/fileutils/files.go
```diff
@@ -128,18 +128,23 @@ func ListFilesRecursiveWalkIntoDirSymlink(path string, walkIntoDirSymlink bool)
 	return
 }
 
-// Return all files with the specified extension in the specified path. Not recursive.
-func ListFilesWithExtension(path, ext string) ([]string, error) {
+// Return all files in the specified path, whose extension satisfy the compareFunc. Not recursive.
+func ListFilesWithExtensionByCompareFunc(path string, compareFunc func(actualExt string) (bool, error)) ([]string, error) {
```
eyalbe4, line 132: I suggest we generalize this function a bit more, and make the filtering function work on the complete file name, not only the extension. This will allow filtering by the extension as well, depending on the logic of the received filtering function.

I also suggest renaming this function from ```ListFilesWithExtensionByCompareFunc``` to ```ListFilesByFilterFunc``` and also rename ```compareFunc func(actualExt string)``` to ```filterFunc func(fileName string)```.

eyalbe4 (changes requested):

PR #227, opened 2020-11-10
Encode Url in GetBuildInfo

RobiNino:


artifactory/services/buildinfo.go
```diff
@@ -49,33 +49,40 @@ func NewBuildInfoParams() BuildInfoParams {
 	return BuildInfoParams{}
 }
 
-// Returns the build info and it's uri of the provided parameters.
-func (bis *BuildInfoService) GetBuildInfo(params BuildInfoParams) (*buildinfo.PublishedBuildInfo, error) {
+// Returns the build info and it's uri of the provided parameters. If build info was not found (404), returns notFound=true.
```
eyalbe4, line 52: it's --> its

artifactory/services/buildinfo.go
```diff
@@ -49,33 +49,40 @@ func NewBuildInfoParams() BuildInfoParams {
...
+func (bis *BuildInfoService) GetBuildInfo(params BuildInfoParams) (pbi *buildinfo.PublishedBuildInfo, notFound bool, err error) {
```
eyalbe4, line 53: I think it  would be nicer to return *found* instead of *notFound*.

eyalbe4 (changes requested):

PR #237, opened 2020-12-02
Remove the go.mod file from test data

talarian1:
-----


utils/io/fileutils/files_test.go
```diff
@@ -44,6 +46,12 @@ func TestGetFileOrDirPathFile(t *testing.T) {
 		assert.Error(t, err)
 		return
 	}
+	// Creates go.mod file
```
eyalbe4, line 49: It is best not to add or remove files to the project sources. The whole operation should be performed in in a temp directory. I believe some of out tests already do this (copy the test project into a temp location, run the tests there. and then remove the file created there).

eyalbe4 (changes requested):

utils/io/fileutils/files_test.go
talarian1: Part of this test is based on the fact that the test data is inside the `jfrog-client-go` project (by assuming a go.mod file will be found somewhere upstream the test data dir). So I don’t see a big benefit with creating all directories over creating a single file.

eyalbe4 (approved):

PR #240, opened 2020-12-07
add support for colored logs

yudkeh:
-----


utils/log/logger.go
```diff
@@ -9,6 +12,13 @@ import (
 var Logger Log
 
 type LevelType int
+type LogFormat string
+
+var Format LogFormat
```
eyalbe4, line 17: This variable seems to be unused, so let's remove it.

utils/log/logger.go
```diff
@@ -40,89 +50,82 @@ type jfrogLogger struct {
 func SetLogger(newLogger Log) {
 	Logger = newLogger
 }
-
 func (logger *jfrogLogger) SetLogLevel(LevelEnum LevelType) {
 	logger.LogLevel = LevelEnum
 }
-
```
eyalbe4, line 47: Let's restore this new-line (we separate functions with a new-line).
This applies to other places in this file.

eyalbe4 (changes requested):
Let's run "go mod tidy", and add the go.mod and/or go.sum if modified.

eyalbe4 (approved):
Something on your local seems to be removing the new-lines :)
I'll push another commit on top of this PR to add them back.

PR #251, opened 2020-12-14
Support including build-dependencies when searching by build

barbelity:
-----


artifactory/services/utils/artifactoryutils.go
```diff
@@ -242,47 +242,73 @@ func createBodyForLatestBuildRequest(buildName, buildNumber string) (body []byte
...
 	}
 
-	wg.Add(1)
-	// Get Sha1 for artifacts by build name and number
 	go func() {
-		buildArtifactsSha1, aqlSearchErr = fetchBuildArtifactsSha1(buildName, buildNumber, flags)
-		wg.Done()
+		// Get Sha1 for artifacts.
+		defer wg.Done()
+		if !specFile.ExcludeArtifacts {
```
eyalbe4, line 260: ExcludeArtifacts --> DepsOnly

artifactory/services/utils/artifactoryutils.go
```diff
@@ -242,47 +242,73 @@ func createBodyForLatestBuildRequest(buildName, buildNumber string) (body []byte
...
+	go func() {
+		// Get Sha1 for dependencies.
+		defer wg.Done()
+		if specFile.IncludeDeps {
+			buildDependenciesSha1, dependenciesAqlSearchErr = fetchBuildArtifactsOrDependenciesSha1(buildName, buildNumber, flags, false)
 		}
-		defer tempReader.Close()
+	}()
+
+	if specFile.ExcludeArtifacts || itemsAlreadyContainProperties {
```
eyalbe4, line 273: This if condition seems wrong, because if itemsAlreadyContainProperties is true, we shouldn't collect the properties.

artifactory/services/utils/specutils.go
```diff
@@ -20,21 +20,23 @@ type ArtifactoryCommonParams struct {
...
+	Exclusions       []string
+	Target           string
+	Props            string
+	ExcludeProps     string
+	SortOrder        string
+	SortBy           []string
+	Offset           int
+	Limit            int
+	Build            string
+	ExcludeArtifacts bool
```
eyalbe4, line 33: ExcludeArtifacts --> DepsOnly

eyalbe4 (changes requested):

artifactory/services/utils/artifactoryutils.go
barbelity: Right, the properties collection happens outside of the 'if' statement.
It reaches the properties collection only if we want to include artifacts, and the properties are missing from the search results.

barbelity:
We discussed about changing the 'exclude-artifacts' flag to 'deps-only', and decided to leave it as 'exclude-artifacts'.
Therefore, I am marking as 'resolved' all the comments requesting this change.

PR #263, opened 2020-12-22
Bugfix - explode fail to extract files with no r permissions 

Or-Geva:
-----

### DIFF
--- a/utils/archiveutils.go
+++ b/utils/archiveutils.go
@@ -23,8 +23,6 @@ func ExtractArchive(localPath, localFileName, originFileName, logMsgPrefix strin
 	if err != nil {
 		return err
 	}
-	// TODO: localFileName may already include the local path.
-	// We should investigate and make sure 'localFileName' contains file name only.
 	var archivePath string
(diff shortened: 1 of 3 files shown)


utils/archiveutils.go
```diff
@@ -45,11 +40,14 @@ func ExtractArchive(localPath, localFileName, originFileName, logMsgPrefix strin
 		return err
 	}
 	log.Info(logMsgPrefix+"Extracting archive:", archivePath, "to", extractionPath)
-	return extract(archivePath, extractionPath)
+	if err = extract(archivePath, originFileName, extractionPath); err != nil {
+		log.Info(logMsgPrefix+"failed to extract", archivePath, ". error:", err.Error())
```
eyalbe4, line 44: If we'd like to add a log message to the error, let's wrap the error with a new one, append the message from the original error to the new one, and return it. We'd like to avoid multiple logging for the same error. We therefore need to avvoid "adding log messages along the way" while errors are thrown upstream the stacktrace.

eyalbe4 (changes requested):

PR #262, opened 2020-12-21
Ptreng 1054 add support for groups

chb0github:
- [x] All [tests](https://github.com/jfrog/jfrog-client-go#tests) passed. If this feature is not already covered by the tests, I added new tests.
- [x] This pull request is on the dev branch.
- [x] I used gofmt for formatting the code before submitting the pull request.
-----

PR #266, opened 2020-12-30
PTRENG-1054 - add support for groups

chb0github:
-----


artifactory/services/groups.go
```diff
@@ -0,0 +1,89 @@
...
+
+type Group struct {
+	Name            string `json:"name,omitempty"`
+	Description     string `json:"description,omitempty"`
+	AutoJoin        bool   `json:"autoJoin,omitempty"`
+	AdminPrivileges bool   `json:"adminPrivileges,omitempty"`
+	Realm           string `json:"realm,omitempty"`
+	RealmAttributes string `json:"realmAttributes,omitempty"`
+}
+type GroupService struct {
```
eyalbe4, line 19: Let's make sure we have a single new line between the declaration of structs and functions.

artifactory/services/groups.go
```diff
@@ -0,0 +1,89 @@
...
+		"Content-Type": "application/json",
+		"Accept":       "application/json",
+	}
+	//
+	url := fmt.Sprintf("%sapi/security/groups/%s", gs.ArtDetails.GetUrl(), group.Name)
+	resp, body, err := gs.client.SendPut(url, content, &httpDetails)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode > 204 {
```
eyalbe4, line 65: See my comments in the closed PR - it is prefered not to check ranges of HTTP response codes. Let's check that the expected value is returned, and return an error otherwise.

artifactory/services/groups.go
```diff
@@ -0,0 +1,89 @@
...
+		return fmt.Errorf("%d %s: %s", resp.StatusCode, resp.Status, string(body))
```
eyalbe4, line 66: See my comment in the closed PR about this - let's stuck to the existing convention.

eyalbe4 (changes requested):
I didn't go over all the code again, because I noticed that some of the comments in https://github.com/jfrog/jfrog-client-go/pull/262 haven't been implemented yet.
Let me know when you'd like me to review this again.

artifactory/services/groups.go
chb0github: Eyal,

I had no desire to do this - except that when I was getting 400 errors `err` was equal to `nil` - I couldn't use err as an indicator. And, the low level client code doesn't consider 400 an error. I will revisit this code.

artifactory/services/groups.go
chb0github: So, I know I made that change - I wonder if `gofmt` did this. Do you have a linter in place? I really don't want to continue to chase spacing issues

artifactory/services/groups.go
chb0github: If you look at this, code, you basically realize that if there is no `err`, only then check the statusCode. 

From what I can tell, the go socket code won't return `err` unless it's like a `SO_TIMEOUT`
See this [sof](https://stackoverflow.com/questions/55210301/error-handling-with-http-newrequest-in-go)
Here are the [official docs](https://golang.org/pkg/net/http/)

I am really forced to check the status code

chb0github: Screen Shot 2020-12-30 at 10 23 19 AM

You have examples in code by your team that do this very thing

chb0github: Here is the [code in question](https://github.com/jfrog/jfrog-client-go/blob/dev/artifactory/services/buildinfo.go#L78)

artifactory/services/groups.go
eyalbe4: We don't have a linter that currently fixes this. I definitely need something like this. We welcome any suggested solutions.

artifactory/services/groups.go
eyalbe4: Can you please replace code like
```
	if resp.StatusCode > 204 {
		return fmt.Errorf("%d %s", resp.StatusCode, resp.Status)
	}
```
with:
```
	if resp.StatusCode != http.StatusOK {
		return fmt.Errorf("%d %s", resp.StatusCode, resp.Status)
	}
```
This is what I meant when suggested not to check respose code ranges. Anything other than what you expect to receive as a successful response code, is an error.

eyalbe4 (changes requested):

artifactory/services/groups.go
chb0github: Sure, I think I can accommodate.

chb0github:
i believe I have made all the changes you've asked. Please let me know if there is anything else

eyalbe4 (changes requested):
The PRs still seem to include http code ranges check:
```
	if res.StatusCode > http.StatusNoContent {
		return nil, fmt.Errorf("%d %s: %s", res.StatusCode, res.Status, string(body))
	}
```

Also the redundant // is still there:
```
	}
	//
	url := fmt.Sprintf("%sapi/security/groups/%s", gs.ArtDetails.GetUrl(), group.Name)
```

Missing new-lines between functions:
```
func (sm *ArtifactoryServicesManagerImp) GetServiceId() (string, error) {
	systemService := services.NewSystemService(sm.client)
	systemService.ArtDetails = sm.config.GetServiceDetails()
	return systemService.GetServiceId()
}
func (sm *ArtifactoryServicesManagerImp) GetGroup(name string) (*services.Group, error) {
	groupService := services.NewGroupService(sm.client)
	groupService.ArtDetails = sm.config.GetServiceDetails()
	return groupService.GetGroup(name)
}
```

Let's move the standard testing method. See the README.

eyalbe4:
@chb0github,
Since https://github.com/jfrog/jfrog-client-go/pull/271 already adds user management support, this PR can probably be closed.

PR #267, opened 2020-12-30
Add GetRemote() for more information

jasonwbarnett:
-----

This PR is ultimately tied to my desire to extend https://github.com/jfrog/terraform-provider-artifactory using this library over https://github.com/atlassian/go-artifactory

I wanted to get some feedback on this type of feature. The generic `Get()` isn't super useful when needing all of the information of a remote repository. I was thinking of implementing this (i.e. `GetRemote()`) and maybe `GetLocal()` and then in the future doing something more. Thoughts? Ideas on how to do this sort of thing better in general?

### DIFF
--- a/artifactory/services/getrepository.go
+++ b/artifactory/services/getrepository.go
@@ -36,6 +36,19 @@ func (grs *GetRepositoryService) Get(repoKey string) (*RepositoryDetails, error)
 	return repoDetails, nil
 }
 
+func (grs *GetRepositoryService) GetRemote(repoKey string) (*RemoteRepositoryDetails, error) {
+	log.Info("Getting repository '" + repoKey + "' details ...")
+	body, err := grs.sendGet(apiRepositories + "/" + repoKey)
+	if err != nil {
+		return nil, err
+	}
+	repoDetails := &RemoteRepositoryDetails{}
+	if err := json.Unmarshal(body, &repoDetails); err != nil {
+		return repoDetails, errorutils.CheckError(err)
+	}
+	return repoDetails, nil
+}
+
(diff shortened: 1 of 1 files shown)


jasonwbarnett:

eyalbe4:
Thanks for proposing this addition @jasonwbarnett.
This is definitely something we were thinking of adding. I suggest however taking a bit of a different approach towards the new API design.
Let's place this new API in the **RemoteRepositoryService** struct. If you look at the README, you'll see that to create a remote repository of type maven, you use **servicesManager.CreateRemoteRepository().Maven(params)**.
We can use the same approach and add **servicesManager.GetRemoteRepository().Maven()**.
The above method will return **MavenRemoteRepositoryParams**, which is already used for creating and updating maven repositories. The same goes for all repository types.
Let me know what you think.

jasonwbarnett:
I'm not a golang developer so I'm not sure what idiomatic API's really look like. I do like the idea of what you're suggesting though. Seems straight forward and allows folks to get exactly what they need.

I had a question around using such an API. How would you expect folks to handle retrieving or iterating over a list of repositories if they didn't know which type of remote repository each one was or whether it was remote or local for that matter? Is there a good way to handle this in golang?

eyalbe4:
Sure @jasonwbarnett,
Go supports interfaces, so any API can return an interface, which can have different possible structs which impelement it. So if you need to store different struct types in the same slice, the slice can be of the type of the interface all structs implement. Another option is to use an [empty interface](https://tour.golang.org/methods/14), which can represent any type.

eyalbe4:
Hi @jasonwbarnett,
Please let us know if you intend to continue promoting this.
Thanks

PR #274, opened 2021-01-25
Support path mapping in distibution commands

yahavi:
-----

Add support for Distribution path mapping.

File spec schema:
```json
{
  "files": [
    {
      "pattern": "[Mandatory] Wildcard pattern match files in the source artifactory",
      "target": "[Optinal] Path in the edge node"
    }
  ]
}

```

The support is builtin in the file spec. When Target is provided, then the path mapping is activated. Artifacts matched in the Pattern will be mapped to the Target after distribution.
To complete the picture, placeholders will be used. The pattern can include capture groups surrounded by parenthesis and the target can include target groups surrounded by curly brackets.

Comparing to the REST API of the [CreateReleaseBundle](https://www.jfrog.com/confluence/display/JFROG/Distribution+REST+API#DistributionRESTAPI-CreateReleaseBundleVersion), there are 2 differences:
1. The pattern is a wildcard and not Regex.
2. The target matching group is the same as [placeholders](https://www.jfrog.com/confluence/display/CLI/CLI+for+JFrog+Artifactory#CLIforJFrogArtifactory-UsingPlaceholders) `{1}`, instead of as Regex `$1`.

### Examples:

---

Example 1:
```go
Pattern: "repo/dir/a.zip"
Target: "repo/dir/b.zip"
```
`repo/dir/a.zip` will be distributed to `repo/dir/b.zip` in the edge node.

---

Example 2:
```go
Pattern: "source-repo/(*)"
Target: "target-repo/{1}"
```
All artifacts at `source-repo` will be distributed to `target-repo`.

---

Example 3:
```go
Pattern: "source-repo/(a)/(*.zip)"
Target: "target-repo/{1}-{2}"
```
For example, an artifact `source-repo/a/bla.zip` will be distributed to `target-repo/a-bla.zip`

### DIFF
--- a/README.md
+++ b/README.md
@@ -805,6 +805,13 @@ params.ReleaseNotes = "Release notes"
 params.ReleaseNotesSyntax = "plain_text"
 params.TargetProps = "key1=val1;key2=val2,val3"
 
+// Artifact path mapping:
+// After distribution, the artifact "source-repo/a/123.zip" destination in the edge node will be "target-repo/a-123.zip".
+// The spec's pattern is a wildcard with capturing groups surrounded by parenthesis. 
+// Respectfully, the spec target's replacement groups surrounded by curly brackets.
+pathMappingSpec := &utils.ArtifactoryCommonParams{Pattern: "source-repo/(a)/(*.zip)", Target: "target-repo/{1}-{2}"}
+params.SpecFiles = append(params.SpecFiles, pathMappingSpec)
+
 err := distManager.CreateReleaseBundle(params)
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -805,6 +805,13 @@ params.ReleaseNotes = "Release notes"
 params.ReleaseNotesSyntax = "plain_text"
 params.TargetProps = "key1=val1;key2=val2,val3"
 
+// Artifact path mapping:
+// After distribution, the artifact "source-repo/a/123.zip" destination in the edge node will be "target-repo/a-123.zip".
+// The spec's pattern is a wildcard with capturing groups surrounded by parenthesis. 
+// Respectfully, the spec target's replacement groups surrounded by curly brackets.
```
eyalbe4, line 811: I suggest modifying the above description to be:

// Be default, artifacts that are distributed as part of a release bundle, have the same path in their destination server 
// (the edge node) as the path they had on the distributing Artifactory server.
// You have however the option for modifying the tagtet path on edge node. You do this by defining the Target property as shown below.
// The Pattern property is a wildcard based pattern. Any wildcards enclosed in parenthesis in the pattern (source) 
// path can be matched with a corresponding placeholder in the target path, to determine the path and name 
// of the artifact, once distributed to the edge node.
// In the following example, the path in the edge node is similar the path in the source Artifactory server, except for the additional "dir" level at the root of the repository.
// Pattern: my-repo/(*)/a.zip
// Target: my-repo/dir/{1}/a.zip

README.md
```diff
@@ -817,6 +824,13 @@ params.ReleaseNotes = "New Release notes"
...
```
eyalbe4, line 830: Let's modify the description to be:

// The Target property defines the target path in the edge node, and can include replaceable in the form of {1}, {2}, ...
// Read more about it in the above "Creating a Release Bundle" section.

eyalbe4 (changes requested):

PR #277, opened 2021-01-31
ant pattern

gailazar300:
-----

### DIFF
--- a/artifactory/services/fspatterns/utils.go
+++ b/artifactory/services/fspatterns/utils.go
@@ -134,9 +134,11 @@ func GetFileSymlinkPath(filePath string) (string, error) {
 
 // Get the local root path, from which to start collecting artifacts to be uploaded to Artifactory.
 // If path dose not exist error will be returned.
-func GetRootPath(pattern, target string, isRegexp, preserveSymLink bool) (string, error) {
+
+//gai
+func GetRootPath(pattern, target string, isRegexp, isAnt, preserveSymLink bool) (string, error) {
 	placeholderParentheses := clientutils.NewParenthesesSlice(pattern, target)
-	rootPath := utils.GetRootPath(pattern, isRegexp, placeholderParentheses)
+	rootPath := utils.GetRootPath(pattern, isRegexp, isAnt, placeholderParentheses)
 	if !fileutils.IsPathExists(rootPath, preserveSymLink) {
 		return "", errorutils.CheckError(errors.New("Path does not exist: " + rootPath))
 	}
--- a/artifactory/services/fspatterns/utils.go
+++ b/artifactory/services/fspatterns/utils.go
@@ -133,7 +133,6 @@ func GetFileSymlinkPath(filePath string) (string, error) {
 
 // Get the local root path, from which to start collecting artifacts to be uploaded to Artifactory.
 // If path dose not exist error will be returned.
-
 func GetRootPath(pattern, target string, patternType clientutils.PatternType, preserveSymLink bool) (string, error) {
 	placeholderParentheses := clientutils.NewParenthesesSlice(pattern, target)
 	rootPath := utils.GetRootPath(pattern, patternType, placeholderParentheses)
--- a/artifactory/services/fspatterns/utils.go
+++ b/artifactory/services/fspatterns/utils.go
@@ -134,7 +134,7 @@ func GetFileSymlinkPath(filePath string) (string, error) {
 // Get the local root path, from which to start collecting artifacts to be uploaded to Artifactory.
 // If path dose not exist error will be returned.
 
-func GetRootPath(pattern, target string, patternType serviceutils.PatternType, preserveSymLink bool) (string, error) {
+func GetRootPath(pattern, target string, patternType clientutils.PatternType, preserveSymLink bool) (string, error) {
 	placeholderParentheses := clientutils.NewParenthesesSlice(pattern, target)
 	rootPath := utils.GetRootPath(pattern, patternType, placeholderParentheses)
 	if !fileutils.IsPathExists(rootPath, preserveSymLink) {
--- a/artifactory/services/fspatterns/utils.go
+++ b/artifactory/services/fspatterns/utils.go
@@ -43,7 +43,7 @@ func PrepareExcludePathPattern(params serviceutils.FileGetter) string {
 		for _, singleExcludePattern := range exclusions {
 			if len(singleExcludePattern) > 0 {
 				singleExcludePattern = utils.ReplaceTildeWithUserHome(singleExcludePattern)
-				singleExcludePattern = utils.PrepareLocalPathForUpload(singleExcludePattern, params.IsRegexp(), params.IsAnt())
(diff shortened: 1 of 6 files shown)


utils/utils.go
```diff
@@ -159,7 +164,7 @@ func CopyMap(src map[string]string) (dst map[string]string) {
 	return
 }
 
-func PrepareLocalPathForUpload(localPath string, useRegExp bool) string {
+func PrepareLocalPathForUpload(localPath string, useRegExp bool, useAnt bool) string {
```
eyalbe4, line 167: I'm not sure it makes sense for this function to receieve two booleans in this case.
It is probably best the function accpets a single argument, which has one of the following 3 values:
1. WildcardPattern
2. RegExp
3. AntPattern
You can define an enum for this. As a reference, look at the ```ItemType``` type we define, to indicate whether we're working with files vs. folders.

The same goes for the ```GetRootPath(pattern, target string, isRegexp, isAnt, preserveSymLink bool) (string, error)``` function signature and a few other functions in the code.

artifactory/services/fspatterns/utils.go
```diff
@@ -133,9 +133,10 @@ func GetFileSymlinkPath(filePath string) (string, error) {
 
 // Get the local root path, from which to start collecting artifacts to be uploaded to Artifactory.
 // If path dose not exist error will be returned.
-func GetRootPath(pattern, target string, isRegexp, preserveSymLink bool) (string, error) {
+
```
eyalbe4, line 136: Redundant new-line.

utils/utils.go
```diff
@@ -169,8 +174,14 @@ func PrepareLocalPathForUpload(localPath string, useRegExp bool) string {
 		localPath = localPath[3:]
 	}
 	if !useRegExp {
-		localPath = pathToRegExp(cleanPath(localPath))
+		if useAnt {
+			localPath = antPathToRegExp(localPath)
+
```
eyalbe4, line 179: Redundant new-line.

utils/utils.go
```diff
@@ -199,6 +210,22 @@ func pathToRegExp(localPath string) string {
 	return "^" + localPath + "$"
 }
 
+func antPathToRegExp(localPath string) string {
+	var SPECIAL_CHARS = []string{".", "^", "$", "+"}
```
eyalbe4, line 214: The go convention does not support camel-casing such as SPECIAL_CHARS.
SPECIAL_CHARS should be changed to **specialChars**.
Tge same goes for the ```wildcardPathToRegExp``` function.

utils/utils.go
```diff
@@ -199,6 +210,22 @@ func pathToRegExp(localPath string) string {
...
+	for _, char := range SPECIAL_CHARS {
+		localPath = strings.Replace(localPath, char, "\\"+char, -1)
+	}
+	var wildcard = ".*"
+	localPath = strings.Replace(localPath, `?`, ".{1}", -1)
+	localPath = strings.Replace(localPath, `**/`, "#", -1)
```
eyalbe4, line 220: The pattern can include **#**, so let's find a way to avoid using it (or any other character) as a temporary replacement.

utils/utils_test.go
```diff
@@ -167,3 +169,65 @@ func TestIsWildcardParentheses(t *testing.T) {
 		t.Errorf("TestIsWildcardParentheses() == %t, want %t", got, want)
 	}
 }
+
+func TestAntPathToRegExp(t *testing.T) {
+	var fileSystemPaths []string = []string{
+		"dev/a/b.txt",
```
eyalbe4, line 175: Let's also add the following path, and see that the logic ant pattern translation works well:
"dev/a1/a2/b.txt"

utils/utils_test.go
```diff
@@ -167,3 +169,65 @@ func TestIsWildcardParentheses(t *testing.T) {
...
+			}
+			var matches []string
+			for _, checkedPath := range fileSystemPaths {
+				match, _ := regexp.MatchString(regExpStr, checkedPath)
+				if match {
+					matches = append(matches, checkedPath)
+				}
+			}
+			if !reflect.DeepEqual(matches, test.matchedPaths) {
+				t.Error("Unmatched! : ant pattern `" + test.antPattern + "` matches urls:\n[" + strings.Join(test.matchedPaths, ",") + "]\nbut got:\n[" + strings.Join(matches, ",") + "]")
```
eyalbe4, line 223: urls --> paths

utils/utils_test.go
```diff
@@ -167,3 +169,65 @@ func TestIsWildcardParentheses(t *testing.T) {
...
+			}
+		})
+
```
eyalbe4, line 226: Redundant new-line.

utils/utils_test.go
```diff
@@ -167,3 +169,65 @@ func TestIsWildcardParentheses(t *testing.T) {
...
+	}
+
```
eyalbe4, line 228: Redundant new-line.

eyalbe4 (changes requested):

gailazar300:
Another review needed.

bintray/services/upload.go
```diff
@@ -240,6 +241,16 @@ func getSingleFileToUpload(rootPath, targetPath string, flat bool) clientutils.A
 	return clientutils.Artifact{LocalPath: rootPathOrig, TargetPath: uploadPath}
 }
 
+func (up *UploadParams) GetPatternType() clientutils.PatternType {
```
eyalbe4, line 244: Notice that thw code of this method is duplicated. See the same method in ```ArtifactoryCommonParams```.
I suggest moving this code to ```github.com/jfrog/jfrog-client-go/utils``` and reusing it.

utils/utils.go
```diff
@@ -199,6 +215,22 @@ func pathToRegExp(localPath string) string {
 	return "^" + localPath + "$"
 }
 
+func antPatternToRegExp(localPath string) string {
+	var specialChars = []string{".", "^", "$", "+"}
+	for _, char := range specialChars {
+		localPath = strings.Replace(localPath, char, "\\"+char, -1)
+	}
```
eyalbe4, line 222: Since the above code line are duplicated in this function and the  ```wildcardPathToRegExp``` function, let's move then into a new private ```replaceSpecialChars``` function, and reuse it.

eyalbe4 (changes requested):

utils/utils_test.go
```diff
@@ -167,3 +169,85 @@ func TestIsWildcardParentheses(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func addRegExpPrefixAndSuffix(str string) string {
+	return "^" + str + "$"
+}
+
+func equalSlicesIgnoreOrder(s1, s2 []string) bool {
```
eyalbe4, line 235: I think we can simplify this function, by sorting first, and then comparing the two slices.
For additional performance, we can compare the size of the slices, before sorting. If the sizes do not match, return false immediately.

eyalbe4 (changes requested):

PR #288, opened 2021-02-11
Add support to git submodules

RobiNino:


utils/tests/utils.go
```diff
@@ -102,3 +104,25 @@ func exitOnErr(err error) {
...
+	assert.NoError(t, err)
+	if found, err := fileutils.IsDirExists(filepath.Join(tmpDir, "gitdata"), false); found {
+		assert.NoError(t, err)
+		err := fileutils.RenamePath(filepath.Join(tmpDir, "gitdata"), filepath.Join(tmpDir, ".git"))
+		assert.NoError(t, err)
+	}
+	submoduleDst := filepath.Join(tmpDir, "subDir", "subModule")
+	err = fileutils.CopyFile(submoduleDst, filepath.Join(tmpDir, "gitSubmoduleData"))
+	assert.NoError(t, err)
+	err = os.Rename(filepath.Join(submoduleDst, "gitSubmoduleData"), filepath.Join(submoduleDst, ".git"))
```
eyalbe4, line 123: I think @Or-Geva was recently faced with an issue with ``` os.Rename```, but can't remember what it was. I think he ended up not using this API.
Worth asking him about it.

utils/vcsdetails.go
```diff
@@ -89,12 +89,12 @@ func (this *VcsCache) clearCacheIfExceedsMax() {
 	}
 }
 
-func tryGetGitDetails(path string, this *VcsCache) (string, string, error) {
-	exists, err := fileutils.IsDirExists(filepath.Join(path, ".git"), false)
+func tryGetGitDetails(path string) (string, string, error) {
```
eyalbe4, line 92: Looks like this fucntion no longer needs to return an error.

eyalbe4 (changes requested):

utils/tests/utils.go
Or-Geva: You are right @eyalbe4.
@RobiNino, let's use [this](https://github.com/jfrog/jfrog-client-go/blob/master/utils/io/fileutils/files.go#L571) instead of `os.Rename`

utils/io/fileutils/files_test.go
```diff
@@ -76,15 +77,12 @@ func TestGetFileOrDirPathFile(t *testing.T) {
...
-		assert.Error(t, err)
+
+	if err = assertFindUpstreamExists(t, "goDotMod.test", projectRoot, File); err != nil {
 		return
 	}
-	assert.True(t, exists, "File goDotMod.test is missing.")
-	assert.Equal(t, projectRoot, root)
 
-	root, exists, err = FindUpstream("go-missing.mod", File)
+	root, exists, err := FindUpstream("go-missing.mod", File)
```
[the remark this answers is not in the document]
RobiNino: Here we assert False & Empty, while `assertFindUpstreamExists` asserts True & Equal.

utils/vcsdetails.go
RobiNino: `extractGitDetails` may still return an error.

PR #298, opened 2021-02-28
Fix IsEncrypted for new Openssl format

yahavi:
-----

In the new OpenSSH format of the private SSH keys, the private keys do not contain the `Proc-Type: ENCRYPTED` header 
(See https://security.stackexchange.com/a/129729). Therefore, we have to find out whether the key is encrypted using another method.

To resolve this issue, I used [ParsePrivateKey](https://pkg.go.dev/golang.org/x/crypto/ssh#ParsePrivateKey) of Go builtin ssh package.

### DIFF
--- a/auth/sshlogin.go
+++ b/auth/sshlogin.go
@@ -135,7 +135,7 @@ func IsEncrypted(buffer []byte) (bool, error) {
 		}
 	}
 	// Key is not encrypted or an error occurred
-	return false, nil
+	return false, err
 }
 
 func parseUrl(url string) (protocol, host string, port int, err error) {
--- a/auth/sshlogin.go
+++ b/auth/sshlogin.go
@@ -3,19 +3,17 @@ package auth
 import (
 	"bytes"
 	"encoding/json"
-	"encoding/pem"
 	"errors"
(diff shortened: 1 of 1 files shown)


auth/sshlogin.go
```diff
@@ -129,11 +127,15 @@ func readSshKeyAndPassphrase(sshKeyPath, sshPassphrase string) ([]byte, []byte,
...
+	_, err := ssh.ParsePrivateKey(buffer)
+	if err != nil {
+		if _, ok := err.(*ssh.PassphraseMissingError); ok {
+			// Key is encrypted
+			return true, nil
+		}
 	}
-	return strings.Contains(block.Headers["Proc-Type"], "ENCRYPTED"), nil
+	// Key is not encrypted or an error occurred
+	return false, nil
```
eyalbe4, line 138: Shouldn't we return the err here? We're always returning nil.

eyalbe4 (changes requested):

auth/sshlogin.go
yahavi: Right

PR #299, opened 2021-03-01
Upload archive

asafgabai:
-----


asafgabai:

asafgabai:

utils/io/content/contentreader.go
```diff
@@ -93,18 +100,25 @@ func (cr *ContentReader) Reset() {
...
-		if err := errorutils.CheckError(os.Remove(cr.filePath)); err != nil {
+	var failedPaths []string
+	for _, filePath := range cr.filesPaths {
+		if filePath == "" {
+			continue
+		}
+		if err := errorutils.CheckError(os.Remove(filePath)); err != nil {
 			log.Error(err)
-			return err
+			failedPaths = append(failedPaths, filePath)
```
eyalbe4, line 110: It looks like we're accumulating this list only for length check at the end.
Can't we just use an int for this?

eyalbe4 (changes requested):
I didn't go through everything yet. I'm releasing what I have so far.
I'll release the rest soon.
(reached up to files.go).

utils/io/fileutils/files.go
```diff
@@ -337,7 +337,30 @@ func calcChecksumDetails(filePath string) (ChecksumDetails, error) {
...
+		defer pw.Close()
+		details.Size, err = io.Copy(pw, reader)
+	}()
+
+	details.Checksum, err = calcChecksumDetailsFromReader(pr)
+	if err != nil {
+		return nil, err
+	}
+
+	return details, nil
```
eyalbe4, line 359: The following code:
```
	if err != nil {
		return nil, err
	}

	return details, nil
```
can be reduced to:
```
	return details, err
```

utils/io/fileutils/files.go
```diff
@@ -337,7 +337,30 @@ func calcChecksumDetails(filePath string) (ChecksumDetails, error) {
...
 	}
-	checksumInfo, err := checksum.Calc(file)
+	return calcChecksumDetailsFromReader(file)
+}
+
+func GetFileDetailsFromReader(reader io.Reader) (*FileDetails, error) {
+	var err error
+	details := new(FileDetails)
+
+	pr, pw := io.Pipe()
```
eyalbe4, line 347: I think that pr should be closed with defer.

eyalbe4 (changes requested):
Let's also update the README with the new / modified APIs.
Let's create a follow up PR with the tests.

utils/io/content/contentreader.go
asafgabai: No, because in case some of the files fail to close, then this list becomes the new filesPaths list.

http/httpclient/client.go
```diff
@@ -185,7 +184,8 @@ func (jc *HttpClient) UploadFile(localPath, url, logMsgPrefix string, httpClient
 	return
 }
 
-func (jc *HttpClient) doUploadFile(localPath, url string, httpClientsDetails httputils.HttpClientDetails, progress ioutils.ProgressMgr) (*http.Response, []byte, error) {
+func (jc *HttpClient) doUploadFile(localPath, url string, httpClientsDetails httputils.HttpClientDetails,
+	progress ioutils.ProgressMgr, progressExtraInfo string) (*http.Response, []byte, error) {
```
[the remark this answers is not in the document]
asafgabai: It was changed so the `ProgressMgr` now gets the full path of the request and it cuts out the props (on the jfrog-cli side). This change will be pushed very soon.

artifactory/services/upload.go
```diff
@@ -124,10 +177,42 @@ func addSymlinkProps(artifact clientutils.Artifact, uploadParams UploadParams) (
 	return clientutils.AddProps(uploadParams.GetTargetProps(), artifactProps), nil
```
eyalbe4, line 177: Let's add a description for the ```addSymlinkProps``` function. Specifically, let's describe the first dtring returned value.

artifactory/services/upload.go
```diff
@@ -124,10 +177,42 @@ func addSymlinkProps(artifact clientutils.Artifact, uploadParams UploadParams) (
...
+		writersMap[data.Artifact.TargetPath].writer.Write(data)
+	}
+}
+
+func collectFilesForUpload(uploadParams UploadParams, progressMgr ioutils.ProgressMgr, vcsCache *clientutils.VcsCache, dataHandlerFunc uploadDataHandlerFunc) error {
 	if strings.Index(uploadParams.GetTarget(), "/") < 0 {
 		uploadParams.SetTarget(uploadParams.GetTarget() + "/")
 	}
+	if uploadParams.Archive != "" && strings.HasSuffix(uploadParams.GetTarget(), "/") {
+		return errors.New("An archive's target cannot be a directory.")
```
eyalbe4, line 214: We should wrap this error with CheckError.

artifactory/services/upload.go
```diff
@@ -159,19 +244,22 @@ func collectFilesForUpload(uploadParams UploadParams, producer parallel.Runner,
...
 		uploadData := UploadData{Artifact: artifact, TargetProps: props, BuildProps: buildProps}
-		task := artifactHandlerFunc(uploadData)
 		if progressMgr != nil {
-			progressMgr.IncGeneralProgressTotalBy(1)
+			if uploadParams.Archive != "" {
+				progressMgr.IncGeneralProgressTotalBy(2)
+			} else {
+				progressMgr.IncGeneralProgressTotalBy(1)
+			}
 		}
```
eyalbe4, line 253: Since the above code is duplicated twice in this file, let's move it into a provate function and resuse it.
Let's also explain why ```IncGeneralProgressTotalBy``` receives 2 in the case of an archive.

eyalbe4 (changes requested):

PR #314, opened 2021-03-15
Implement GetUserTokens lookup function

sebinjohn:
-----

Implements two changes
1. Change the anonymous struct inside GetTokensResponseData.
2. Implement a GetUserTokens lookup function to lookup tokens issued to a username.

I would like to implement some tests for the second item, but require some help around what is the best way to write a test case for function which doesn't take *Params as args.

### DIFF
--- a/tests/artifactorysecurity_test.go
+++ b/tests/artifactorysecurity_test.go
@@ -148,6 +148,9 @@ func getUserTokenstest(t *testing.T) {
 	if tokens[0] != token.AccessToken {
 		t.Error("Retried user token doesn't match expected token value")
 	}
+
+	tokensToRevoke = append(tokensToRevoke, token.RefreshToken)
+
 	params := services.NewCreateTokenParams()
 	params.Username = "test-user"
 	params.Scope = "api:* member-of-groups:readers"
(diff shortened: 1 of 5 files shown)


eyalbe4 (changes requested):
Thanks for adding this @sebinjohn.
Let's add a test for this and also document the new API in the README.

sebinjohn:
> Thanks for adding this @sebinjohn.
> Let's add a test for this and also document the new API in the README.

@eyalbe4 I would like to implement some tests for the second item, but require some help around what is the best way to write a test case for function which doesn't take *Params as args.

eyalbe4:
@sebinjohn, I think you can take a similar approach to the one taken in ```getTokensTest```,

sebinjohn:
@eyalbe4  I have implemented a test case.

tests/artifactorysecurity_test.go
```diff
@@ -136,6 +136,42 @@ func getTokensTest(t *testing.T) {
 	tokensToRevoke = append(tokensToRevoke, token.RefreshToken)
 }
 
+func getUserTokenstest(t *testing.T) {
```
eyalbe4, line 139: Let's make sure that the test revokes the tokens it generates. See the ```tokensToRevoke``` slice used in other tokens tests.

eyalbe4 (changes requested):

tests/artifactorysecurity_test.go
sebinjohn: Done.

PR #338, opened 2021-04-25
Fix release bundle creation with dry-run

RobiNino:
-----

### DIFF
--- a/tests/distribution_test.go
+++ b/tests/distribution_test.go
@@ -177,9 +177,16 @@ func updateDryRun(updateBundleParams services.UpdateReleaseBundleParams) error {
 	return testsBundleUpdateService.UpdateReleaseBundle(updateBundleParams)
 }
 
+func distributeDryRun(distributionParams services.DistributionParams) error {
+	defer setServicesToDryRunFalse()
+	testsBundleDistributeService.DryRun = true
+	return testsBundleDistributeService.Distribute(distributionParams)
+}
+
 func setServicesToDryRunFalse() {
 	testsBundleCreateService.DryRun = false
 	testsBundleUpdateService.DryRun = false
+	testsBundleDistributeService.DryRun = false
 }
 
 func createWithProps(t *testing.T) {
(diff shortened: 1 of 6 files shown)


tests/distribution_test.go
```diff
@@ -107,33 +106,73 @@ func createUpdate(t *testing.T) {
...
+	// Create release bundle params
 	createBundleParams := services.NewCreateReleaseBundleParams(bundleName, bundleVersion)
 	createBundleParams.Description = "Release bundle description 1"
 	createBundleParams.ReleaseNotes = "Release notes 1"
 	createBundleParams.SpecFiles = []*utils.ArtifactoryCommonParams{{Pattern: RtTargetRepo + "b.in"}}
-	err := testsBundleCreateService.CreateReleaseBundle(createBundleParams)
-	assert.NoError(t, err)
+
+	// Test DryRun first
+	err := createDryRun(createBundleParams)
```
eyalbe4, line 116: Since this is a dry-run, we can validate that the release bundle was not created, in addition to the err check we have now.
The same goes for the ```updateDryRun``` action - we can check that the bundle is not modified after the operation.

tests/distribution_test.go
```diff
@@ -172,9 +211,13 @@ func createWithProps(t *testing.T) {
 	// Check prop1Values and prop2Values
 	assert.Len(t, prop1Values, 1)
 	assert.Len(t, prop2Values, 2)
-	assert.Equal(t, "value1", prop1Values[0])
-	assert.Equal(t, "value2", prop2Values[0])
-	assert.Equal(t, "value3", prop2Values[1])
+	if len(prop1Values) == 1 {
```
eyalbe4, line 214: What is the reason for this len check? We do an asset of this len a few lines above.
Same goes for ```if len(prop2Values) == 2``` below.

eyalbe4 (changes requested):

tests/distribution_test.go
RobiNino: The assertion doesn't fail immediately, so the length check prevents a NPE

eyalbe4 (approved):

PR #344, opened 2021-04-29
remove omitempty since default is true

matteosilv:
-----

FailFast needs to be always printed in the response since it defaults to true

### DIFF
--- a/artifactory/services/promote.go
+++ b/artifactory/services/promote.go
@@ -85,15 +85,16 @@ func (ps *PromoteService) BuildPromote(promotionParams PromotionParams) error {
 }
 
 type BuildPromotionBody struct {
-	Comment             string              `json:"comment,omitempty"`
-	SourceRepo          string              `json:"sourceRepo,omitempty"`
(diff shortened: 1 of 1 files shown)


yahavi (changes requested):
Thanks for your contribution, @matteosilv!
Could you please add a comment above `FailFast` explaining why we don't have `omitempty` in FailFast? It isn't trivial at all.

matteosilv:
Hi @yahavi, without omitempty the value will be always marshalled whether it is true or false. I have a pull request jfrog/jfrog-cli#1070, and basing on existing code of both projects I saw that other fields with true default value are handled this way. In other terms it seems to me that is in charge of jfrog-cli (in these particular case) of handling default values.

yahavi:
@matteosilv 
Yes, It took me some minutes to understand why this change is needed. Therefore I think a comment in the code above `FailFast` is mandatory.

yahavi (approved):
Thanks @matteosilv!

PR #349, opened 2021-05-11
Add sha256 detailed-summary to release bundle

gailazar300:
-----


distribution/services/create.go
```diff
@@ -41,11 +42,17 @@ func (cb *CreateReleaseBundleService) CreateReleaseBundle(createBundleParams Cre
...
 }
 
-func (cb *CreateReleaseBundleService) execCreateReleaseBundle(gpgPassphrase string, releaseBundle *createReleaseBundleBody) error {
+// In case of an immediate sign- release bundle detailed summary (containing sha256) will be returned.
+// In other cases summary will be nil.
+func (cb *CreateReleaseBundleService) execCreateReleaseBundle(gpgPassphrase string, releaseBundle *createReleaseBundleBody) (*clientutils.Sha256Summary, error) {
+	var summary *clientutils.Sha256Summary = nil
+	if releaseBundle.SignImmediately {
+		summary = clientutils.NewSha256Summary()
+	}
```
yahavi, line 51: I suggest doing exactly as you did in the [buildinfo.go](https://github.com/jfrog/jfrog-client-go/blob/dev/artifactory/services/buildinfo.go#L74): Initialize the summary in the beginning (with falsy default `succeeded` value) and set `succeeded=true` and `sha256` in the end.
This way I think the code will be shorter and `succeeded` will be false in case of a failure.
Same goes for the update command.
```suggestion
	var summary *clientutils.Sha256Summary = clientutils.NewSha256Summary()
```

yahavi (changes requested):

PR #354, opened 2021-05-26
feat: use Build info to create release bundle

sguiheux:
I noticed that ArtifactoryCommonParams.Build is never use during the creation of a release bundle.
I use the "same" code as Bundle parameter. Works fine :)

### DIFF
--- a/artifactory/services/utils/aqlquerybuilder.go
+++ b/artifactory/services/utils/aqlquerybuilder.go
@@ -23,12 +23,16 @@ func CreateAqlBodyForSpecWithPattern(params *ArtifactoryCommonParams) (string, e
 	itemTypeQuery := buildItemTypeQueryPart(params)
 	nePath := buildNePathPart(triplesSize == 0 || includeRoot)
 	excludeQuery := buildExcludeQueryPart(params, triplesSize == 0 || params.Recursive, params.Recursive)
+	buildQueryPart, err := buildBuildInfoQuery(params)
+	if err != nil {
+		return "", err
+	}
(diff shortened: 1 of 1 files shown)


yahavi (changes requested):
@sguiheux 
Thanks for this pull request!
The `CreateAqlBodyForSpecWithPattern` function is widely used by other use cases. For example, when you run a search command with pattern and build, for performance-wise, we run 2 AQL queries - the one in `CreateAqlBodyForSpecWithPattern`,  and [another one](https://github.com/jfrog/jfrog-client-go/blob/v0.23.1/artifactory/services/utils/aqlquerybuilder.go#L103) to filter the results by the input build. 
If you want, you can track the flow of [SearchBySpecWithPattern](https://github.com/jfrog/jfrog-client-go/blob/v0.23.1/artifactory/services/utils/searchutil.go#L161).

Here you have a special use case whereby you have no choice but to put all filters in only 1 AQL. Therefore I think we should be careful when adding the build Artifacts to this AQL and make sure that only in the release bundles specific use case they will be added to the AQL generated by `CreateAqlBodyForSpecWithPattern`. 

Please let me know if there is anything unclear.

sguiheux:
Hi, sorry for the delay.
I understand, but not sure to know how to fix this :/

sguiheux:
It's strange to have a method to create a something from an input, and you cannot use all the input because it can break something.
Maybe the input is not the good one, and we must have 2 functions no?

eyalbe4:
@sguiheux - 
We unintentionally abandoned this discussion a few minths ago.
Would you like to share some info about the use case? Are you still interested in promoting this?

PR #378, opened 2021-06-30
Placeholders logic ignores '--flat'

Or-Geva:
-----
Placeholders define the target path of each file for upload/download/move/copy commands. As `--flat` makes sense while using wildcard patter, there is no concrete meaning for using it with placeholders, resulting in unintuitive results.
For example, `mypath/{1}/{2}/{3}/{1}' as a target path, ` --flat` has no meaning since we create the structure for each file, As a result, the question if to use ` --flat` makes no sense here.

Resolves: https://github.com/jfrog/jfrog-cli/issues/883
Related PR: https://github.com/jfrog/jfrog-cli/pull/1131

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/jfrog/jfrog-client-go/utils/io/fileutils"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


utils/utils.go
```diff
@@ -275,18 +277,27 @@ func BuildTargetPath(pattern, path, target string, ignoreRepo bool) (string, err
...
+	if len(groups) > 0 {
+		target, replaceOccurred := ReplacePlaceHolders(groups, target)
+		return target, replaceOccurred, nil
+	}
+	return target, false, nil
+}
+
+// group - regular expression matched group to replace with placeholders
+// toReplace - target pattern to replace
+func ReplacePlaceHolders(groups []string, toReplace string) (string, bool) {
```
eyalbe4, line 293: Let's also describe the returned values.

eyalbe4 (changes requested):

PR #385, opened 2021-07-08
Allow disabling checksum calculation before uploading

eyalbe4:
-----
```
Add the option of disabling all checksum calculation prior to uploading a file to Artifactory.
Here's how the checksum calculation can be disabled:
params := services.NewUploadParams()
...
// The min file size in bytes for "checksum deploy".
// "checksum deploy" is the action of calculating the file checksum locally, before 
// the upload, and skipping the actual file transfer if the file already  
// exists in Artifactory.
// MinChecksumDeploy default value: 10400
params.MinChecksumDeploy = 15360
// Set to false to disable all checksum calculation, including "checksum deploy".
// ChecksumsCalcEnabled default value: true
params.ChecksumsCalcEnabled = false 

totalUploaded, totalFailed, err := rtManager.UploadFiles(params)
```


asafgabai (approved):
Just one minor change is needed

PR #402, opened 2021-08-01
Add response body prints to retry warn messages

talarian1:
-----

### DIFF
--- a/http/httpclient/client.go
+++ b/http/httpclient/client.go
@@ -81,11 +81,13 @@ func (jc *HttpClient) SendPut(url string, content []byte, httpClientsDetails htt
 	return
 }
 
-func (jc *HttpClient) newRequest(method, url string, body io.Reader) (*http.Request, error) {
+func (jc *HttpClient) newRequest(method, url string, body io.Reader) (req *http.Request, err error) {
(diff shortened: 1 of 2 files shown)


http/httpclient/client.go
```diff
@@ -127,6 +125,15 @@ func (jc *HttpClient) Send(method, url string, content []byte, followRedirect, c
 	return
 }
 
+func (jc *HttpClient) createReq(method, url string, content []byte) (req *http.Request, err error) {
+	log.Debug(fmt.Sprintf("Sending HTTP %s request to: %s", method, url))
+	if content != nil {
+		return jc.newRequest(method, url, bytes.NewBuffer(content))
+	}
+	req, err = jc.newRequest(method, url, nil)
+	return req, errorutils.CheckError(err)
```
eyalbe4, line 134: The CheckError should probably not be here, but moved into the ```jc.newRequest``` method.
Note that due to the fact that the error isn't checked inside ```jc.newRequest```, the error returned in line 131 is never checked.
After we make the above change, we should also look at all the usages of ```jc.newRequest``` and remove the CheckError's from them.

eyalbe4 (changes requested):

PR #415, opened 2021-08-16
Fix: Upload with archive and include-dirs leaves out empty directories

gailazar300:
-----
Uploading a folder as a zip file with --include-dirs will exclude any empty directories in the folder.

### DIFF
--- a/artifactory/services/upload.go
+++ b/artifactory/services/upload.go
@@ -776,9 +776,15 @@ func (us *UploadService) addFileToZip(artifact *clientutils.Artifact, progressPr
 	if errorutils.CheckError(e) != nil {
 		return
 	}
+	// Check if symlink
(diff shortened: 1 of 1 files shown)


artifactory/services/upload.go
```diff
@@ -775,6 +775,12 @@ func (us *UploadService) addFileToZip(artifact *clientutils.Artifact, progressPr
 		_, e = writer.Write([]byte(filepath.ToSlash(artifact.SymlinkTargetPath)))
 		return
 	}
+	// Check if directory
+	// Write directory to writer - path needs to end with a separator.
```
eyalbe4, line 779: // Check if directory
// Write directory to writer - path needs to end with a separator.
-->
// If this is a directory, add it to the writer with a trailing slash.

eyalbe4 (changes requested):

PR #434, opened 2021-09-26
Fix exclusions in transitive search

asafgabai:
- [x] All [tests](https://github.com/jfrog/jfrog-client-go#tests) passed. If this feature is not already covered by the tests, I added new tests.
- [x] This pull request is on the dev branch.
- [x] I used gofmt for formatting the code before submitting the pull request.
-----
Exclusion patterns must start with a repository name or an asterisk.
Fixed a bug when an exclusion started with an asterisk in transitive search, which caused an error to be returned from Artifactory.

PR #435, opened 2021-09-26
Fix exclusions in transitive search (v1)

asafgabai:
-----
Exclusion patterns must start with a repository name or an asterisk.
Fixed a bug when an exclusion started with an asterisk in transitive search, which caused an error to be returned from Artifactory.


eyalbe4 (changes requested):
See my comment for https://github.com/jfrog/jfrog-client-go/pull/434

Also, would you mind also adding a clear description to this PR?
This is important for future tracabloty, and also makes it easier for everyone to understand the intent of this fix, before reading the code.

PR #440, opened 2021-10-03
Bugfix: unzip an archive uploaded with "Archive" and "IncludeDirs" parameters doesn't work

gailazar300:
-----
Unzip a zip file uploaded with the "UploadFiles" function with "IncludeDirs" and "Archive" parameters- using the os "unzip" command doesn't work. The empty directories are extracted as files and not as directories.
There is no problem using the "DownloadFiles" function with the "Explode" parameter.


artifactory/services/upload.go
```diff
@@ -765,6 +765,13 @@ func (us *UploadService) addFileToZip(artifact *clientutils.Artifact, progressPr
 		header.Name = clientutils.TrimPath(localPath)
 	}
 	header.Method = zip.Deflate
+
+	// If this is a directory, add it to the writer with a trailing slash.
+	if info.IsDir() {
+		header.Name = header.Name + "/"
```
eyalbe4, line 771: Can be shortened to - 
```
header.Name += "/"
```
Also, do we always need to add the slash? Can the name sometimes end with a slash?

PR #449, opened 2021-10-14
Fix: Fails to collect buildinfo vcs information for repository without commits

gailazar300:
-----
For example: fails after running upload command "jfrog rt u" with buildinfo flags "--build-name=.." "--build-number=..." and when the local directory contains a repository without commits.

### DIFF
--- a/utils/git.go
+++ b/utils/git.go
@@ -44,7 +44,9 @@ func (m *manager) ReadConfig() error {
 	m.handleSubmoduleIfNeeded()
 	m.readRevisionAndBranch()
 	m.readUrl()
-	m.readMessage()
+	if m.revision != "" {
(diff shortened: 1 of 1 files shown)


utils/git.go
```diff
@@ -276,8 +278,6 @@ func (m *manager) readRevisionFromPackedRef(ref string) {
 			return
 		}
 	}
-
-	m.err = errorutils.CheckError(errors.New("failed fetching revision from git config, from ref: " + ref))
```
RobiNino, line 280: I suggest adding a log message instead of this error. Something like "No packed-refs file was found. Assuming git repository is empty"

RobiNino (changes requested):

PR #450, opened 2021-10-17
Replace Appveyor with GitHub Actions

Or-Geva:
-----


tests/utils_test.go
```diff
@@ -38,90 +39,97 @@ import (
...
+	testsPipelinesIntegrationsService *pipelinesServices.IntegrationsService
+	testsPipelinesSourcesService      *pipelinesServices.SourcesService
+
+	// Access Services
+	testsAccessProjectService *accessServices.ProjectService
+
+	timestamp       = time.Now().Unix()
+	trueValue       = true
+	falseValue      = false
+	randomRunNumber = strings.Replace(uuid.NewString(), "-", "", -1)
```
yahavi, line 116: This is not a number. What do you think about `runUid`?
```suggestion
	runUid = strings.Replace(uuid.NewString(), "-", "", -1)
```

yahavi (changes requested):

tests/pipelinessources_test.go
```diff
@@ -16,6 +18,10 @@ func testAddPipelineSource(t *testing.T) {
 		assert.NotEmpty(t, *PipelinesVcsToken, "cannot run pipelines tests without vcs token configured")
 		return
 	}
+	if utils.IsMacOS() {
+		// The following test modify a single Pipelines resource. Therefore, we limit the parallelism by running on MacOS (tests run on all OS parallel).
+		t.Skip("Not running on MacOS, skipping...")
+	}
```
yahavi, line 24: I think this is problematic. The tests should be allowed to run on all operating systems (unless there is a limitation in the test, such as Docker on Windows). Issues in the CI level should not impact the tests.
However, unless I'm missing something, I don't see any limitation on the CI level to run in parallel. the `integrationName` is random. 
```suggestion
```

PR #451, opened 2021-10-17
CreateFileTransferDetails - prevent double slash in target path

talarian1:
-----

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -5,7 +5,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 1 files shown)


utils/utils.go
```diff
@@ -497,9 +499,12 @@ type DeployableArtifactDetails struct {
 	TargetRepository string `json:"targetRepository,omitempty"`
 }
 
-func (detailes *DeployableArtifactDetails) CreateFileTransferDetails(rtUrl, targetRepository string) FileTransferDetails {
-	targetPath := rtUrl + targetRepository + "/" + detailes.ArtifactDest
-	return FileTransferDetails{SourcePath: detailes.SourcePath, TargetPath: targetPath, Sha256: detailes.Sha256}
+func (details *DeployableArtifactDetails) CreateFileTransferDetails(rtUrl, targetRepository string) (FileTransferDetails, error) {
+	url, err := url.Parse(rtUrl + targetRepository)
```
eyalbe4, line 503: We should and return this error here, if it is not nil.
I suggest we also explain why we're parsing the URL.

PR #456, opened 2021-10-28
Sanitize windows paths better on unarchive

yahavi:
-----

To prevent absolute target paths outside of an extraction directory, we should use `filepath` instead of `path`.
This issue appeared after creating `TestUploadWithArchiveAndSymlinkZipSlip` in https://github.com/jfrog/jfrog-cli/pull/1269/files#diff-90b1ac3745fa656b3bbf876ff9778354050d926be2c38f2909ac328ef9542c1bR2017 and running it on Windows.

### DIFF
--- a/utils/io/fileutils/archive.go
+++ b/utils/io/fileutils/archive.go
@@ -182,7 +182,7 @@ func isEntryInDestination(destinationDir, pathInArchive string) bool {
 	if os.IsPathSeparator('\\') && strings.HasPrefix(pathInArchive, "/") {
 		return false
 	}
-	// Since the entry in archive should be always represented as Unix path, the "path" module is used and not "filepath"
+
 	pathInArchive = filepath.Clean(pathInArchive)
 	if !filepath.IsAbs(pathInArchive) {
 		// If path is relative, concatenate it to the destination dir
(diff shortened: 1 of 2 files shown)


utils/io/fileutils/archive.go
```diff
@@ -166,18 +168,25 @@ func checkSymlinkEntry(header *archiveHeader, archiveEntry archiver.File, destin
...
 }
 
 // Make sure the extraction path of the archive entry is under the destination dir
 func isEntryInDestination(destinationDir, pathInArchive string) bool {
+	// If pathInArchive starts with '/' and we are on Windows, the path is illegal
+	pathInArchive = strings.TrimSpace(pathInArchive)
+	if os.IsPathSeparator('\\') && strings.HasPrefix(pathInArchive, "/") {
+		return false
+	}
 	// Since the entry in archive should be always represented as Unix path, the "path" module is used and not "filepath"
```
eyalbe4, line 185: Looks like this comment is no longer valid, following the replacement of ```path``` with ```filepath```. Is this correct?

eyalbe4 (changes requested):

PR #494, opened 2021-12-20
Add a retryWaitTime option when enabling retry mechanism

gailazar300:
-----
Adding a new 'retryWaitTime' field that sets the number of seconds/milli-seconds to sleep between retries for upload, download, move, copy, delete and set/delete properties commands.


eyalbe4 (changes requested):
Looks like the ```RetriesInterval``` is described as *Number of seconds ...* when it should actually be *Number of milliseconds ...*.
```
type RetryExecutor struct {
	// The amount of retries to perform.
	MaxRetries int

	// Number of seconds to sleep between retries.
	RetriesInterval int

	// Message to display when retrying.
	ErrorMessage string
```

I also suggest renaming ```RetriesInterval``` to ```RetriesIntervalMilliSecs```.
In addition, to constant and clearer, let's rename all across the code
```SetRetryWaitTime``` --> ```SetRetryWaitMilliSecs```
```GetHttpRetryWaitTime``` --> ```GetHttpRetryWaitMilliSecs```
```httpRetryWaitTime``` --> ```httpRetryWaitMilliSecs```
```GetHttpRetryWaitTime``` --> ```GetHttpRetryWaitMilliSecs```
```retryWaitTime``` --> ```retryWaitMilliSecs```

File: utils/retryexecutor.go [hunk 1]
```diff
@@ -38,9 +39,10 @@ func (runner *RetryExecutor) Execute() error {
 		}
 
 		log.Warn(runner.getLogRetryMessage(i, err))
-		// Going to sleep for RetryInterval seconds
+		// Going to sleep for RetryInterval milliseconds
 		if runner.RetriesInterval > 0 && i < runner.MaxRetries {
-			time.Sleep(time.Second * time.Duration(runner.RetriesInterval))
+			log.Info("Waiting " + strconv.Itoa(runner.RetriesInterval) + "ms before trying again")
```
eyalbe4, line 44: Can be simplified to - 
```
log.Info("Waiting", runner.RetriesInterval, "ms before trying again")
```
Notice that using a comma between arguments also takes care fo the spacing between sections of log message.

utils/retryexecutor_test.go
```diff
@@ -13,7 +14,7 @@ func TestRetryExecutorSuccess(t *testing.T) {
 
 	executor := RetryExecutor{
 		MaxRetries:      retriesToPerform,
-		RetriesInterval: 0,
+		RetriesInterval: 5000,
```
eyalbe4, line 17: Is this change actually necessary? Wouldn't it make the test no longer, with no additional testing value?

eyalbe4 (changes requested):

PR #503, opened 2022-01-06
Determine target path inside an uploaded archive using placeholders

gailazar300:
-----
When using the 'archive' option for upload, we can control the target path inside the uploaded archive using placeholders.


eyalbe4 (changes requested):
I'm really not sure we should have a ```TerraformService``` as part of jfrog-client-go, for the same reason we don't have a service for publishing npm, pip or go packages. This logic should be handled by jfrog-cli-core.
I'll wait for you to make these changes, before I continue reviewing this PR, as well as the other dependent PRs pn jfrog-cli-core, and jfrog-cli, since I assume the changes will also reflect on the other PRs.

eyalbe4 (changes requested):
Will you be able to rename this PR, so that it describes what it does?
If needed, worth also adding a clear description for it.
I believe that following the changes you added to it, its name and description need to be modified.

artifactory/services/upload.go
```diff
@@ -344,8 +348,9 @@ func createUploadTask(taskData *uploadTaskData, dataHandlerFunc uploadDataHandle
 	} else {
 		taskData.target = getUploadTarget(symlinkPath, taskData.target, taskData.uploadParams.IsFlat(), placeholdersUsed)
 	}
+	targetPathInArchive, _ := clientutils.ReplacePlaceHolders(taskData.groups, taskData.uploadParams.TargetPathInArchive)
```
eyalbe4, line 351: Let's describe this line.

artifactory/services/upload.go
```diff
@@ -588,6 +593,8 @@ type UploadParams struct {
 	MinChecksumDeploy    int64
 	ChecksumsCalcEnabled bool
 	Archive              string
+	// When using 'archive' option we can set the target path in the archive using placeholders.
```
eyalbe4, line 596: I'm not sure the above description is clear enough. How about - 

// When using the 'archive' option for upload, we can control the target path inside the uploaded archive using placeholders. This operation determines the TargetPathInArchive value.

http/httpclient/client.go
```diff
@@ -271,6 +271,9 @@ func (jc *HttpClient) UploadFileFromReader(reader io.Reader, url string, httpCli
 	if errorutils.CheckError(err) != nil {
 		return nil, nil, err
 	}
+	if resp != nil && resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusOK {
+		return nil, nil, errorutils.CheckError(errors.New("Artifactory response: " + resp.Status))
```
eyalbe4, line 275: Since this is a generic http client, and not an http client used for Artifactory only, let's change - 
errors.New("Artifactory response: " + resp.Status)
-->
errors.New("Server response: " + resp.Status)

eyalbe4 (changes requested):

PR #505, opened 2022-01-11
Upload a file more than once

asafgabai:
-----
When walking the file tree, the same file might be visited more than once, but a dir symlink won't be visited more than once.


utils/io/fileutils/listfiles.go
```diff
@@ -81,11 +91,23 @@ func walk(path string, info os.FileInfo, walkFn WalkFunc, visited map[string]boo
...
 	if err != nil {
 		return walkFn(root, nil, err)
 	}
-	return walk(root, info, walkFn, visited, walkIntoDirSymlink)
+	return walk(root, info, walkFn, visitedDirSymlinks, walkIntoDirSymlink)
+}
+
+// Gets a path of a file or a directory, and returns its real path (in case the path contains a symlink to a directory).
+// The difference between this function and filepath.EvalSymlinks is that if the path is of a symlink,
+// the function won't return the symlink's target, but the real path to the symlink.
```
eyalbe4, line 103: "the function" - it's not clear from reading this sentence, which of the two functions discussed in the description, "the function" refers to. It should be probably changed to "this function" or "filepath.EvalSymlinks".

eyalbe4 (changes requested):

PR #513, opened 2022-01-26
V1 - Remove 'build-info published' log message

eyalbe4:
-----
This is part of a fix for V1, which has already been implemented for V2 a few months ago.
This PR removes the "Build info successfully published" log message. It is a duplicate of https://github.com/jfrog/jfrog-client-go/pull/419 which has been merged to v2 a few months ago. This was done, in order to support a fix for an issue, whereby the build-info link displayed in the log, was incorrect, when the build is published as part of a project. 
As part of the fix, the fixed link is built and displayed by jfrog-cli-core, and is therefore removed from jfrog-client-go in this PR.
A PR for jfrog-cli-core will be soon created.

### DIFF
--- a/auth/cert/sslutils_default.go
+++ b/auth/cert/sslutils_default.go
@@ -1,3 +1,4 @@
+//go:build !windows
 // +build !windows
 
 package cert
--- a/auth/cert/sslutils_windows.go
+++ b/auth/cert/sslutils_windows.go
@@ -1,3 +1,4 @@
+//go:build windows
 // +build windows
 
 package cert
--- a/artifactory/services/buildinfo.go
+++ b/artifactory/services/buildinfo.go
@@ -3,9 +3,6 @@ package services
 import (
 	"encoding/json"
 	"errors"
-	"net/http"
-	"net/url"
-
 	"github.com/jfrog/jfrog-client-go/artifactory/buildinfo"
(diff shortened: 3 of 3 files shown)


auth/cert/sslutils_default.go
```diff
@@ -1,3 +1,4 @@
+//go:build !windows
```
RobiNino, line 1: Is this relevant to the PR?

auth/cert/sslutils_windows.go
```diff
@@ -1,3 +1,4 @@
+//go:build windows
```
RobiNino, line 1: Is this relevant to the PR?

RobiNino (changes requested):

auth/cert/sslutils_default.go
eyalbe4: It is not, but this is something go 1.17 automatically modifies. It did the same for v2 as well, after we started releasing it with go 1.17.
I'll make sure that v1 is also moved to go 1.17.

auth/cert/sslutils_windows.go
eyalbe4: It is not, but this is something go 1.17 automatically modifies. It did the same for v2 as well, after we started releasing it with go 1.17.
I'll make sure that v1 is also moved to go 1.17.

PR #515, opened 2022-01-31
Delete default test repo after repositories tests

RobiNino:
Added missing test type to teardown.

### DIFF
--- a/tests/utils_test.go
+++ b/tests/utils_test.go
@@ -48,7 +48,7 @@ var (
 	TestXray                 *bool
 	TestPipelines            *bool
 	TestAccess               *bool
-	TestRepository           *bool
+	TestRepositories         *bool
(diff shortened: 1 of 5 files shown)


tests/utils_test.go
```diff
@@ -606,7 +602,7 @@ func createRepo() error {
 }
 
 func teardownIntegrationTests() {
-	if !(*TestArtifactory || *TestDistribution || *TestXray) {
+	if !(*TestArtifactory || *TestDistribution || *TestXray || *TestRepository) {
```
eyalbe4, line 605: I think we should rename ```TestRepository``` to ```TestRepositories```, becuase plural makes more sense.

eyalbe4 (changes requested):

PR #518, opened 2022-01-31
Build promotion failFast error logs

sverdlov93:
-----
When running build promotion with failFast=false the command runs despite any errors that happen during the promotion.
Added a log to print the errors.

### DIFF
--- a/artifactory/services/promote.go
+++ b/artifactory/services/promote.go
@@ -92,8 +92,7 @@ type BuildPromotionBody struct {
 	Status              string `json:"status,omitempty"`
 	IncludeDependencies *bool  `json:"dependencies,omitempty"`
 	Copy                *bool  `json:"copy,omitempty"`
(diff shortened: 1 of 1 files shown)


artifactory/services/promote.go
```diff
@@ -91,6 +93,7 @@ type BuildPromotionBody struct {
 	IncludeDependencies *bool  `json:"dependencies,omitempty"`
 	Copy                *bool  `json:"copy,omitempty"`
 	// FailFast options default is true. We need to avoid omitempty, otherwise, it would be forced to false if omitted.
+	// Notice that when running 'JFrog CLI V1' FailFast's default is false.
```
eyalbe4, line 96: jfrog-client-go is not supposed to be aware or includes information about the applications that use it. It is not jfrog-cli's client, but an independent library. I therefore suggest removign this line.

eyalbe4 (changes requested):

PR #524, opened 2022-02-14
Fixes #523 Add capability to specify proxy for repository replications

### LINKED ISSUE #523: Add capability to specify proxy for Artifactory repository replications
**Is your feature request related to a problem? Please describe.**
I have a use case where I am trying to configure Artifactory replications through Terraform and need to be able to set a proxy for the replication.

**Describe the solution you'd like to see**
Add support for setting the proxy when creating a new replication.  After support is added in this repo, changes will also be needed for https://github.com/jfrog/terraform-provider-artifactory.

**Describe alternatives you've considered**
N/A

**Additional context**
* https://github.com/atlassian/go-artifactory/pull/33 tried adding support for it when `terraform-provider-artifactory` was using that repo for the client.  We were using that until we upgraded to Artifactory 7.x where the JSON API for GET and POST replications was slightly changed.

Guent4:
-----

I have a use case where I am trying to configure Artifactory replications through Terraform and need to be able to set a proxy for the replication.  Couple things to note about the current changes:
1. The GET JSON response for getting a replication has changed between Artifactory 6.x and 7.x such that this change would only work for 7.x.  See https://github.com/atlassian/go-artifactory/pull/33, an old PR,  for the change for Artifactory 6.x.
2. I wanted to added a test for this, but unfortunately was not able to figure out how to dynamically create a proxy via a REST API (couldn't find anything on https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API).  If you can help point me in the right direction wrt this, I would love to add a test for configuring proxy.  I have tested the changes using a GUI/manually created proxy.

### DIFF
--- a/artifactory/services/utils/replication.go
+++ b/artifactory/services/utils/replication.go
@@ -1,11 +1,14 @@
 package utils
 
+import "encoding/json"
+
 type ReplicationBody struct {
 	Username               string `json:"username"`
 	Password               string `json:"password"`
 	URL                    string `json:"url"`
 	CronExp                string `json:"cronExp"`
 	RepoKey                string `json:"repoKey"`
+	Proxy                  string `json:"proxy"`
 	EnableEventReplication bool   `json:"enableEventReplication"`
 	SocketTimeoutMillis    int    `json:"socketTimeoutMillis"`
 	Enabled                bool   `json:"enabled"`
@@ -22,6 +25,7 @@ type ReplicationParams struct {
 	CronExp  string
 	// Source replication repository.
 	RepoKey                  string
+	Proxy                    string
 	EnableEventReplication   bool
 	SocketTimeoutMillis      int
 	Enabled                  bool
@@ -32,6 +36,27 @@ type ReplicationParams struct {
 	IncludePathPrefixPattern string
 }
 
+// UnmarshalJSON overrides the default JSON unmarshal function because the POST request to create a replication
+// has a field named `proxy` but the GET request returns a JSON with a field named `proxyRef`
+func (rp *ReplicationParams) UnmarshalJSON(data []byte) error {
+	type Alias ReplicationParams
+
+	aux := &struct {
+		ProxyRef string `json:"proxyRef"`
+		*Alias
+	}{
+		Alias: (*Alias)(rp),
+	}
+
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	rp.Proxy = aux.ProxyRef
+
+	return nil
+}
+
 func CreateReplicationBody(params ReplicationParams) *ReplicationBody {
 	return &ReplicationBody{
 		Username:               params.Username,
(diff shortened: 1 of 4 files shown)


Guent4:

artifactory/services/utils/replication.go
```diff
@@ -32,13 +36,35 @@ type ReplicationParams struct {
 	IncludePathPrefixPattern string
 }
 
+// UnmarshalJSON overrides the default JSON unmarshal function because the POST request to create a replication
+// has a field named `proxy` but the GET request returns a JSON with a field named `proxyRef`
+func (rp *ReplicationParams) UnmarshalJSON(data []byte) error {
```
eyalbe4, line 41: I suggest taking a different approach instead of defining the ```UnmarshalJSON``` function.
We can do the following:
1. Remove the ```Proxy``` property from the ```ReplicationBody``` struct.
2. Create two new structs named ```GetReplicationBody``` and ```UpdateReplicationBody```.
3. The two structs should reference the ```ReplicationBody``` struct, and only add the ```Proxy``` or ```ProxyRef``` properties.

This way, the fields are replicated na the ```ReplicationBody``` struct is used as a base for ```GetReplicationBody``` and ```UpdateReplicationBody```.

Guent4: Thank you for the suggestion!  I have made the changes accordingly.

Guent4:
The force push was because I had accidentally committed the fix using a different user and needed to re-write that commit.

@eyalbe4 I noticed that there was an xray test that was failing from your test trigger.  Do you have any insights into what might be causing that?  Thanks in advance!

artifactory/services/utils/replication.go
```diff
@@ -1,14 +1,11 @@
 package utils
 
-import "encoding/json"
-
-type ReplicationBody struct {
+type replicationBody struct {
```
Guent4, line 3: One downside of this change to have two new structs of `GetReplicationBody` and `UpdateReplicationBody` rather than overriding the JSONUnmarshal function is that this change is no longer backwards compatible.

Even if I keep choose to keep `replicationBody` exposed, the function specs in [artifactory/services/createreplication.go](https://github.com/jfrog/jfrog-client-go/pull/524/commits/e7c1168ef582b47998ad992e2ba1b7aecc45a366#diff-bc8309bdb397edd2769ac91efbcfc86a9935ca146a9c83bcee9666440a306edb) and [artifactory/services/getreplication.go](https://github.com/jfrog/jfrog-client-go/pull/524/commits/e7c1168ef582b47998ad992e2ba1b7aecc45a366#diff-0f25ef08a33a2041ce11c417ad456871009fd7bcd6b427719f678e5fe06113d4) will be changed which are backward incompatible changes.

eyalbe4: @Guent4 -
We're mostly okay with breaking compatibility following APO changes. If a public API changed, we make sure to note it in the release notes.

artifactory/services/utils/replication.go
cjconrad2: @eyalbe4, just a curious consumer here. Would you mind explaining why you prefer your approach to the original proposed change? Seems like the initial change would have been backwards compatible. I see your later comment about being generally okay with backwards incompatible changes, but if there is a way to do something in a non-breaking way, why not take that approach? Just trying to understand rationale here, not trying to be a pain!

eyalbe4:
Thanks so much for this important addition @Guent4.
It will be released shortly.

PR #537, opened 2022-03-03
Add static analysis

RobiNino:
-----
Run a Github action with static analysis and gosec scan on every pr/push (parallel to tests).
Both tools can also be run from the command line.
A few rules are temporarily ignored and will be fixed separately.


yahavi (changes requested):

.github/workflows/analysis.yml
```diff
@@ -0,0 +1,24 @@
+name: "Static Analysis"
+on: ["push", "pull_request"]
+jobs:
+  Static-Check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v2
+      - name: Static Code Analysis
+        uses: dominikh/staticcheck-action@v1.1.0
```
yahavi, line 10: We can use the latest minor
```suggestion
        uses: dominikh/staticcheck-action@v1
```

.github/workflows/analysis.yml
```diff
@@ -0,0 +1,24 @@
...
+        with:
+          version: "2021.1.2"
```
yahavi, line 12: Can we use the latest?
```suggestion
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,5 @@
 - [ ] All [tests](https://github.com/jfrog/jfrog-client-go#tests) passed. If this feature is not already covered by the tests, I added new tests.
+- [ ] All [static analysis checks](https://github.com/jfrog/jfrog-client-go/actions/workflows/analysis.yml) passed.
```
yahavi, line 2: I'd prefer adding instructions for running the static checks in the readme, like we do for the tests. 
```suggestion
- [ ] All [static analysis checks](https://github.com/jfrog/jfrog-client-go#static-check) passed.
```

RobiNino: Will add in the future as discussed

auth/sshlogin.go
```diff
@@ -128,11 +129,9 @@ func readSshKeyAndPassphrase(sshKeyPath, sshPassphrase string) ([]byte, []byte,
 
 func IsEncrypted(buffer []byte) (bool, error) {
 	_, err := ssh.ParsePrivateKey(buffer)
-	if err != nil {
```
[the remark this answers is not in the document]
RobiNino: Yep, staticcheck:
`S1020  when ok is true, err can't be nil`

.github/workflows/analysis.yml
RobiNino: `Because new versions of Staticcheck introduce new checks that may break your build, it is recommended to pin to a specific version and to update Staticheck consciously.`
They recommend otherwise, but we can try latest and pin a version if it breaks too much.

Issue #528, opened 2022-02-17
Support Xray Artifact Summary `xray/api/v2/summary/artifact`


tenstad:
**Is your feature request related to a problem? Please describe.**
There is no support for fetching artifact summaries.  
https://www.jfrog.com/confluence/display/XRAY2X/Xray+REST+API#XrayRESTAPI-ArtifactSummary

**Describe the solution you'd like to see**
Implement `GetArtifactSummary`, similar to [`GetBuildSummary`](https://github.com/jfrog/jfrog-client-go/blob/85c68385bc2264ff426b92958b6fb515fd0ab07f/xray/services/summary.go#L32).

PR #544, opened 2022-03-22
Implement GetArtifactSummary

tenstad:
-----

Closes https://github.com/jfrog/jfrog-client-go/issues/528

### DIFF
--- a/xray/services/summary.go
+++ b/xray/services/summary.go
@@ -94,10 +94,9 @@ type ArtifactSummaryResponse struct {
 }
 
 type Artifact struct {
-	General General `json:"general,omitempty"`
-	Issues  []Issue `json:"issues,omitempty"`
-	// TODO: Create License struct with correct fields for api/v2/summary/artifact endpoint
-	// Licenses []License `json:"licenses,omitempty"`
+	General  General          `json:"general,omitempty"`
+	Issues   []Issue          `json:"issues,omitempty"`
+	Licenses []SummaryLicense `json:"licenses,omitempty"`
 }
 
 type General struct {
@@ -124,6 +123,13 @@ type Issue struct {
 	Created     string             `json:"created,omitempty"`
 	ImpactPath  []string           `json:"impact_path,omitempty"`
 	Components  []SummaryComponent `json:"components,omitempty"`
+}
+
+type SummaryLicense struct {
+	Components  []string `json:"components,omitempty"`
+	FullName    string   `json:"full_name,omitempty"`
+	MoreInfoUrl []string `json:"more_info_url,omitempty"`
+	Name        string   `json:"name,omitempty"`
 }
 
 type Error struct {
--- a/xray/services/summary.go
+++ b/xray/services/summary.go
@@ -3,11 +3,13 @@ package services
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
+
+	servicesutils "github.com/jfrog/jfrog-client-go/artifactory/services/utils"
 	"github.com/jfrog/jfrog-client-go/auth"
 	"github.com/jfrog/jfrog-client-go/http/jfroghttpclient"
 	"github.com/jfrog/jfrog-client-go/utils"
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
-	"net/http"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


tenstad:

eyalbe4:
recheckcla

xray/services/summary.go
```diff
@@ -53,6 +55,59 @@ func (ss *SummaryService) GetBuildSummary(params XrayBuildParams) (*SummaryRespo
...
+type ArtifactSummaryResponse struct {
+	Artifacts []Artifact `json:"artifacts,omitempty"`
+	Errors    []Error    `json:"errors,omitempty"`
+}
+
+type Artifact struct {
+	General General `json:"general,omitempty"`
+	Issues  []Issue `json:"issues,omitempty"`
+	// TODO: Create License struct with correct fields for api/v2/summary/artifact endpoint
+	// Licenses []License `json:"licenses,omitempty"`
```
tenstad, line 100: A `License` struct already exists https://github.com/jfrog/jfrog-client-go/blob/e3c6206b724288677c125d89268f461a78c33e57/xray/services/scan.go#L225-L231, but does not match the api specifiaction of https://www.jfrog.com/confluence/display/XRAY2X/Xray+REST+API#XrayRESTAPI-ArtifactSummary .

Is the existing one correct, and should I simply create a new `ArtifactLicense` struct?

eyalbe4:
Thanks for this contribution @tenstad!
Will you be able to add the new API to the README?

tenstad:
> Thanks for this contribution @tenstad! Will you be able to add the new API to the README?

Ofc! Added a new section to the README now 😊

We could skip the first two lines where `summaryService` is created. I included them because `GetArtifactSummary` requires a `SummaryService` with `XrayDetails`, which I do not find intuitive to create. Maybe I am doing something wrong.

tenstad:
> which I do not find intuitive to create. Maybe I am doing something wrong.

Ah, it is because I did not implement a wrapper method on `XrayServiceManager`. I missed this because `GetBuildSummary` does not have such a method.

Fixed for `GetArtifactSummary` now, and I suggest that `GetBuildSummary` is added to the README and a corresponding wrapper is created for it in another PR.

xray/services/summary.go
mikaello: The existing [`License` struct](https://github.com/jfrog/jfrog-client-go/blob/e3c6206b724288677c125d89268f461a78c33e57/xray/services/scan.go#L225-L231) belongs to the `/scan/graph` API it seems, when comparing to the TypeScript library and Java library they have created two different datatypes for the license used in scan graph summary and in summary:

*TypeScript*:
* [Scan/GraphLicense.ts](https://github.com/jfrog/jfrog-client-js/blob/master/model/Xray/Scan/GraphLicense.ts)
* [Summary/License.ts](https://github.com/jfrog/jfrog-client-js/blob/master/model/Xray/Summary/License.ts)

*Java*:
* [scan/License.java](https://github.com/jfrog/xray-client-java/blob/master/src/main/java/com/jfrog/xray/client/services/scan/License.java)
* [summary/License.java](https://github.com/jfrog/xray-client-java/blob/master/src/main/java/com/jfrog/xray/client/services/summary/License.java)

So I think it is safe to say that you should create a new license struct.

xray/services/summary.go
```diff
@@ -53,6 +55,59 @@ func (ss *SummaryService) GetBuildSummary(params XrayBuildParams) (*SummaryRespo
...
+}
+
+type General struct {
+	ComponentId string `json:"component_id,omitempty"`
+	Name        string `json:"name,omitempty"`
+	Path        string `json:"path,omitempty"`
+	PkgType     string `json:"pkg_type,omitempty"`
+	Sha256      string `json:"sha256,omitempty"`
+}
```
mikaello, line 109: When comparing to [TypeScript](https://github.com/jfrog/jfrog-client-js/blob/master/model/Xray/Summary/General.ts) and [Java](https://github.com/jfrog/xray-client-java/blob/master/src/main/java/com/jfrog/xray/client/services/summary/General.java), you are missing the following properties:

* `Sha1`
* `ParentSha256`

I notice that they are not part of [the documentation](https://www.jfrog.com/confluence/display/XRAY2X/Xray+REST+API#XrayRESTAPI-ArtifactSummary), but I guess the Go lib should document the same API as TypeScript and Java?!

xray/services/summary.go
eyalbe4: Right - please go ahead and create it. We rally prefer not to merge PRs with // TODO comments or commented out code @mikaello .

eyalbe4 (changes requested):
Let's also add a test for new ```GetArtifactSummary``` API. The test can performed using a mock server. You can look at the ```TestNewXrayScanService``` as a reference and do something similar.
You can put the test inside the ```summary_test.go``` file under the ```tests``` directory. You'll see all the test files there.

tenstad:
@eyalbe4 I added a simple test now :) It can definitely be improved/extended in the future, but at least a basic response with all fields is now tested.

Had to add two fields that were missing in existing structs (ComponentPhysicalPaths and Cwe), so that the entire sample response from xray could be retured by the client.

PR #572, opened 2022-05-18
Fix git execution function wrong path assignment

sverdlov93:
Signed-off-by: sverdlov93 <sverdlov93@gmail.com>

-----
On the git execution function, the `Dir` field, which indicates where the execution is running, was assigned mistakenly as the `.git` path.
This caused the git execution calls to run inside the` .git` folder and not the project folder.


eyalbe4 (changes requested):
Let's improve the PR title and description, so that the random reader can understand the issue, by just reading the release notes. Besides that, the code change looks good.

PR #593, opened 2022-06-02
Remove repository name from artifacts' path field in the build-info JSON

Or-Geva:
-----

### DIFF
--- a/artifactory/services/utils/resultutils.go
+++ b/artifactory/services/utils/resultutils.go
@@ -51,7 +51,7 @@ func (cs *OperationSummary) Close() error {
 	return cs.ArtifactsDetailsReader.Close()
 }
 
-func (ad *ArtifactDetails) ToBuildInfoArtifact() buildinfo.Artifact {
+func (ad *ArtifactDetails) ToBuildInfoArtifact() (buildinfo.Artifact, error) {
 	artifact := buildinfo.Artifact{Checksum: buildinfo.Checksum{}}
 	artifact.Sha1 = ad.Checksums.Sha1
 	artifact.Md5 = ad.Checksums.Md5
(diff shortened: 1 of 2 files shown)


artifactory/services/utils/resultutils.go
```diff
@@ -53,13 +53,20 @@ func (ad *ArtifactDetails) ToBuildInfoArtifact() buildinfo.Artifact {
...
+	artifact.Sha256 = ad.Checksums.Sha256
 	// Artifact name in build info as the name in artifactory
 	filename, _ := fileutils.GetFileAndDirFromPath(ad.ArtifactoryPath)
 	artifact.Name = filename
 	if i := strings.LastIndex(filename, "."); i != -1 {
 		artifact.Type = filename[i+1:]
 	}
-	artifact.Path = ad.ArtifactoryPath
+	// Remove the repository from Artifact's path.
+	// Since a build can be promoted, the build-info may be incorrect (builds are immutable).
```
eyalbe4, line 64: I think it's best to simplify the above to - 

// The 'path' property in the build-info should not include the repository. We therefore remove the repository from the path.

artifactory/services/utils/resultutils.go
```diff
@@ -53,13 +53,20 @@ func (ad *ArtifactDetails) ToBuildInfoArtifact() buildinfo.Artifact {
...
+	if i := strings.Index(ad.ArtifactoryPath, "/"); i != -1 {
+		artifact.Path = ad.ArtifactoryPath[i+1:]
+	} else {
```
eyalbe4, line 67: Is there a known case where the path at this stage does not include the repository? If there is, let's explain it in a comment here. If there isn't, let's not use a "defensive" approach, and return an error with a clear "unexpected" message inside this else.

eyalbe4 (changes requested):

eyalbe4 (changes requested):
Let's add "Breaking change" to the title of this PR.

PR #605, opened 2022-06-23
Add git call to find .git directory for git worktrees

Robert-Morgowicz:
Adds a new section to handleSubmoduleIfNeeded which queries git for the git dir if it's in a worktree instead of failing to find the config.  

-----

closes https://github.com/jfrog/jfrog-cli/issues/1531


Robert-Morgowicz:

eyalbe4:
@Robert-Morgowicz,
I see that you're using `exec.Command("git", "rev-parse", "--git-dir")` as a fallback, but that brings the following question - 
What if the code used only this git command in the first place? Wouldn't it always work? Do we still need the code that runs before it? 
If not, will you be able to share more information about the two cases?

eyalbe4:
Got it @Robert-Morgowicz.
In that case. I think we better have only the git command approach, and remove the older code. The code would be simpler and more predicable. We should also see iof there's a test that covers this, and if not, add the test.

Robert-Morgowicz:
> Got it @Robert-Morgowicz. In that case. I think we better have only the git command approach, and remove the older code. The code would be simpler and more predicable. We should also see iof there's a test that covers this, and if not, add the test.

I reduced it to just using the new implementation (asking git for the location of the .git dir).  I tested it and it works for submodules as well.  I couldn't figure out how to add a test for git worktrees, though.  I see where to add one in `utils/vcsdetails_test.go`, but digging into `utils/tests/utils.go`, I don't see where the actual git calls used to create these test repos are coming from.

eyalbe4:
Thank you @Robert-Morgowicz.
Here's how you can testing for the submodules functionality you added.
The `handleSubmoduleIfNeeded` function you improved is called from the `ReadConfig` function,
image

The `ReadConfig` function is tested by the `testReadConfig` function.
image

If you look at the first line of this test function, you'll see that it calls the `getDotGitPath(t)` function, which returns a path to a .git directory for the testing. This .git directory is copied to from the `testdata` directory.
You can take the same approach, and create a new `testReadConfigWithSubmodules` function, which worls similarly like the existing `testReadConfig` function. You can place your own submodules .git inside the tests data, and your new test function can copy and test it test it.

lex-ibm:
recheck

lex-ibm:
Hello @eyalbe4!

We've added a test case for worktrees. Is there anything else we need to do?

lex-ibm:
Ping

eyalbe4:
Let me review this soon @lex-ibm.

notoriousUdo:
Any updates here?

utils/git.go
```diff
@@ -73,38 +74,33 @@ func (m *GitManager) handleSubmoduleIfNeeded() {
...
 		return
 	}
-
-	// Extract path by removing prefix.
-	actualRelativePath := strings.TrimSpace(line[strings.Index(line, ":")+1:])
-	actualAbsPath := filepath.Join(filepath.Dir(m.path), actualRelativePath)
-	exists, err = fileutils.IsDirExists(actualAbsPath, false)
+	ResolvedGitPath := strings.TrimSpace(stdout.String())
+	// trim the worktree path to just the
+	// worktreePathReduced := filepath.Dir(filepath.Dir(worktreePath))
```
eyalbe4, line 92: The above two commented lines seem unclear / unnecessary.

utils/git.go
```diff
@@ -73,38 +74,33 @@ func (m *GitManager) handleSubmoduleIfNeeded() {
...
-		m.err = errorutils.CheckErrorf("failed to parse .git path for submodule")
+		m.err = err
 		return
 	}
-
-	// Extract path by removing prefix.
-	actualRelativePath := strings.TrimSpace(line[strings.Index(line, ":")+1:])
-	actualAbsPath := filepath.Join(filepath.Dir(m.path), actualRelativePath)
-	exists, err = fileutils.IsDirExists(actualAbsPath, false)
+	ResolvedGitPath := strings.TrimSpace(stdout.String())
```
eyalbe4, line 90: ResolvedGitPath --> resolvedGitPath

eyalbe4 (changes requested):

utils/git.go
lex-ibm: Refactored that variable name

lex-ibm:
Hey @eyalbe4! 
Happy holidays!

schmidtl25:
Hi @eyalbe4 - can this PR be re-reviewed and merged?

PR #624, opened 2022-08-03
Improve server get version bad response log

sverdlov93:
-----
Add prefix to getVersion unmarshal error logs: 

For example:
Screen Shot 2022-08-03 at 18 53 35

To this:
Screen Shot 2022-08-03 at 18 50 18


artifactory/services/system.go
```diff
@@ -45,7 +45,7 @@ func (ss *SystemService) GetVersion() (string, error) {
 	var version artifactoryVersion
 	err = json.Unmarshal(body, &version)
 	if err != nil {
-		return "", errorutils.CheckError(err)
```
talarian1, line 48: Should we also print the `body` (as string)?

talarian1 (approved):

PR #630, opened 2022-08-10
Fix flaky StorageInfo test

SteveNewson:
-----
The storage info test fails occasionally because of a race between the
refresh and the subsequent call to fetch the storage information. In
some cases, the storage refresh has not completed before the request to
fetch the storage information occurs.

This change adds a loop with a timeout to the test to wait until the
repository is listed in the storage information before continuing to
assert the test statements.


yahavi (approved):
Thanks for this contribution, @SteveNewson!

Issue #549, opened 2022-03-31
For ease of testing, Properties struct should encode values in predictable/sorted order.


AnalogJ:
**Is your feature request related to a problem? Please describe.**
Our test framework uses Go-VCR which allows us to run our integration test suite without interacting with a real Artifactory server. We've noticed that our tests will occassionally fail because a "recording" could not be found. This is because the generated URL can be different between runs:

```
https://artifactory.corp.example.com/artifactory/myrepository/test/archive.tar.gz;upload_path=test%2Farchive.tar.gz;repo_org=git;repo_name=test":

https://artifactory.corp.example.com/artifactory/myrepository/test/archive.tar.gz;repo_name=test;repo_org=git;upload_path=test%2Farchive.tar.gz"
```

**Describe the solution you'd like to see**
The `ToEncodedString()` function should always generate a string in a predictable/sorted order.
https://github.com/jfrog/jfrog-client-go/blob/e3c6206b724288677c125d89268f461a78c33e57/artifactory/services/utils/properties.go#L97

SteveNewson:
@yahavi This can now be closed.

PR #631, opened 2022-08-11
Encode properties in predictable order

SteveNewson:
-----

When testing, it is useful to be able to have some predictability in the
way that properties are encoded to a string. Since the properties struct
is currently backed by a map, the order that it is rendered to a string
is not predictable.

This change sorts the keys before rendering to a string. The items in
the arrays (the values in the map) are considered to be ordered and
hence are not sorted.

Feature Request: https://github.com/jfrog/jfrog-client-go/issues/549

### DIFF
--- a/artifactory/services/utils/properties.go
+++ b/artifactory/services/utils/properties.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 	"net/url"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


artifactory/services/utils/properties.go
```diff
@@ -95,10 +96,21 @@ func (props *Properties) AddProperty(key, value string) {
...
 func (props *Properties) ToEncodedString(concatValues bool) string {
+	// Sort the keys to create a predictable order
+	keys := make([]string, 0, len(props.properties))
+	for k := range props.properties {
+		keys = append(keys, k)
+	}
+
+	// sort the slice by keys
+	sort.Strings(keys)
+
```
yahavi, line 107: To improve readability, I suggest exporting this logic to a local function: `func (props *Properties) getSortedKeys() []string`.

yahavi (changes requested):
Thanks for this contribution, @SteveNewson!
Please consider my small comment.

yahavi (approved):
Thanks, @SteveNewson!

PR #647, opened 2022-09-04
Allow stopping retry executor

yahavi:
-----

Allow providing the context to the retry executor to allow canceling.

### DIFF
--- a/utils/retryexecutor.go
+++ b/utils/retryexecutor.go
@@ -1,14 +1,20 @@
 package utils
 
 import (
+	"context"
+	"errors"
 	"fmt"
-	"github.com/jfrog/jfrog-client-go/utils/log"
 	"time"
+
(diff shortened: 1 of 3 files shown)


utils/retryexecutor.go
```diff
@@ -63,5 +69,15 @@ func (runner *RetryExecutor) LogRetry(attemptNumber int, err error) {
 	} else {
 		log.Debug(message)
 	}
+}
 
+func (runner *RetryExecutor) isCancelled() bool {
```
eyalbe4, line 74: I suggest having this function return the context cancellation error instead of a boolean. That would allow the retry executor to return the actual error that caused it stop.
I therefoer also suggest renaming this function to `checkCancelled`.

eyalbe4 (changes requested):

PR #649, opened 2022-09-12
Transfer files - Report transfer files usage

omerzi:
-----

### DIFF
--- a/artifactory/usage/reportusage.go
+++ b/artifactory/usage/reportusage.go
@@ -20,6 +20,10 @@ type ReportUsageAttribute struct {
 	AttributeValue string
 }
 
+func (rua *ReportUsageAttribute) isEmpty() bool {
+	return rua.AttributeName == ""
+}
+
 func SendReportUsage(productId, commandName string, serviceManager artifactory.ArtifactoryServicesManager, attributes ...ReportUsageAttribute) error {
 	config := serviceManager.GetConfig()
 	if config == nil {
@@ -74,9 +78,9 @@ func isVersionCompatible(artifactoryVersion string) bool {
 func reportUsageToJson(productId, commandName string, attributes ...ReportUsageAttribute) ([]byte, error) {
 	featureInfo := feature{FeatureId: commandName}
 	if len(attributes) > 0 {
-		featureInfo.Attributes = make(map[string]string, 0)
+		featureInfo.Attributes = make(map[string]string, len(attributes))
 		for _, attribute := range attributes {
-			if attribute != (ReportUsageAttribute{}) {
+			if !attribute.isEmpty() {
 				featureInfo.Attributes[attribute.AttributeName] = attribute.AttributeValue
 			}
 		}
--- a/artifactory/usage/reportusage.go
+++ b/artifactory/usage/reportusage.go
@@ -2,21 +2,20 @@ package usage
 
 import (
 	"encoding/json"
-	"fmt"
-	"net/http"
-
 	"errors"
+	"fmt"
 	versionutil "github.com/jfrog/gofrog/version"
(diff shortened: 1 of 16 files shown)


artifactory/usage/reportusage.go
```diff
@@ -67,8 +66,16 @@ func isVersionCompatible(artifactoryVersion string) bool {
 	return version.AtLeast(minArtifactoryVersion)
 }
 
-func reportUsageToJson(productId, commandName string) ([]byte, error) {
+func reportUsageToJson(productId, commandName, serviceId, serverSize string) ([]byte, error) {
 	featureInfo := feature{FeatureId: commandName}
+	if serviceId != "" && serverSize != "" {
+		attributes := map[string]string{
```
eyalbe4, line 72: * The `SendReportUsage` API in jfrog-client-go should remain generic, and not be aware of the specific needs of the `transfer-files` usage report. It therefore should not be aware of attributes like  `serviceId` and `serverSize`. I suggest this API will receieve an optional attributes map.
* I suggest remaning "serverSize" to "sourceStorageSize".

eyalbe4 (changes requested):

artifactory/usage/reportusage.go
omerzi: Done

artifactory/usage/reportusage.go
```diff
@@ -67,8 +71,16 @@ func isVersionCompatible(artifactoryVersion string) bool {
 	return version.AtLeast(minArtifactoryVersion)
 }
 
-func reportUsageToJson(productId, commandName string) ([]byte, error) {
+func reportUsageToJson(productId, commandName string, attributes ...ReportUsageAttribute) ([]byte, error) {
 	featureInfo := feature{FeatureId: commandName}
+	if len(attributes) > 0 {
+		featureInfo.Attributes = make(map[string]string, 0)
```
yahavi, line 77: It is a good practice to set the capacity, if possible:
```suggestion
		featureInfo.Attributes = make(map[string]string, len(attributes))
```

artifactory/usage/reportusage.go
```diff
@@ -67,8 +71,16 @@ func isVersionCompatible(artifactoryVersion string) bool {
...
+		for _, attribute := range attributes {
+			if attribute != (ReportUsageAttribute{}) {
```
yahavi, line 79: Since you're checking for emptiness multiple times in the code, I suggest creating a method `isEmpty` for readability. Also, there is no meaning in values without a key. Therefore I think it is enough to create and use isEmpty method as follows:
```go
func (rua *ReportUsageAttribute) isEmpty() bool {
	return rua.AttributeName == ""
}
```

artifactory/usage/reportusage_test.go
```diff
@@ -35,34 +36,47 @@ func TestReportUsageJson(t *testing.T) {
 	type test struct {
```
yahavi, line 36: Let's give it an informative name
```suggestion
	type reportUsageTestCase struct {
```

yahavi (changes requested):

yahavi:
Please label this PR before merging

artifactory/usage/reportusage.go
omerzi: done

artifactory/usage/reportusage.go
omerzi: done

artifactory/usage/reportusage_test.go
omerzi: done

yahavi: Let's also move this struct outside of this function to improve readability

PR #674, opened 2022-11-10
Improve reference tokens handling

sverdlov93:
-----


auth/authutils.go
```diff
@@ -76,24 +73,37 @@ func setAudienceManually(tokenPayload *TokenPayload, payload []byte) error {
...
 
-func ExtractUsernameFromAccessToken(token string) (string, error) {
+func ExtractUsernameFromAccessToken(token string) (username string) {
+	var err error
+	defer func() {
+		if err != nil {
+			log.Warn(err.Error() + "\n" +
+				"The provided access token is not a valid JWT, probably a reference token.\n" +
+				"Some build tools only support basic authentication which requires also a username.\n" +
+				"If you plan to work with one of those tools, please provide a username.")
```
eyalbe4, line 83: I like the above warning message, however, I suggest that instead of adding it as a warn message to the log, we will add this message to the existing error and return it.

eyalbe4 (changes requested):

PR #675, opened 2022-11-13
Git LFS Fix - annotation tags unrecognized

omerzi:
-----
- In case of annotated tags, commit object fetch failed.
- Fixed by fetching the annotated tag object, and then fetching the commit from the target hash inside it.


artifactory/services/gitlfsclean.go
```diff
@@ -214,6 +218,21 @@ func getLfsFilesFromGit(path, refMatch string) (map[string]struct{}, error) {
...
+	annotatedTag, err := repo.TagObject(ref.Hash())
+	if err != nil {
+		return nil, errorutils.CheckError(err)
+	}
+	// Find the commit object associated with the annotated tag.
+	commit, err := repo.CommitObject(annotatedTag.Target)
+	if err != nil {
+		return nil, errorutils.CheckError(err)
+	}
+	return commit, err
```
eyalbe4, line 233: Can be simplified to just one line - 
```
	return commit, errorutils.CheckError(err)

```

eyalbe4 (changes requested):

artifactory/services/gitlfsclean.go
omerzi: done

PR #719, opened 2023-03-29
Add FlattenGraph for GraphNode

omerzi:
-----


xray/services/scan.go
```diff
@@ -189,6 +190,34 @@ type GraphNode struct {
...
+func FlattenGraph(graph []*GraphNode) []*GraphNode {
+	flatGraph := []*GraphNode{}
+	depsSet := datastructures.MakeSet[string]()
+	for _, node := range graph {
+		flatGraph = append(flatGraph, getUniqueDependencies(node, depsSet)...)
+	}
+	return []*GraphNode{{Id: "root", Nodes: flatGraph}}
+}
+
+func getUniqueDependencies(node *GraphNode, depsSet *datastructures.Set[string]) []*GraphNode {
```
yahavi, line 203: Wouldn't work better with a map of string -> *GraphNode?
The key is the dependency ID and the value is the GraphNode. No return value.

Possible implementation (please double check me):
```go
func FlattenGraph(graph []*GraphNode) []*GraphNode {
	allDependencies := map[string]*GraphNode{}
	for _, node := range graph {
		populateUniqueDependencies(node, allDependencies)
	}
	return []*GraphNode{{Id: "root", Nodes: maps.Values(allDependencies)}}
}

func populateUniqueDependencies(node *GraphNode, allDependencies map[string]*GraphNode) {
	if _, exist := allDependencies[node.Id]; exist {
		return
	}
	allDependencies[node.Id] = &GraphNode{Id: node.Id}
	for _, dependency := range node.Nodes {
		populateUniqueDependencies(dependency, allDependencies)
	}
}
```

xray/services/scan_test.go
```diff
@@ -47,3 +49,32 @@ func TestCreateScanGraphQueryParams(t *testing.T) {
 		})
 	}
 }
+
+func TestFlattenGraph(t *testing.T) {
```
yahavi, line 53: Let's add another test to check no duplications

yahavi (changes requested):

xray/services/scan.go
omerzi: Nice! Didn't know about the `maps.Values()`. Works great.

xray/services/scan_test.go
omerzi: Already checked it in the test (with the set)

eyalbe4 (approved):

PR #743, opened 2023-04-16
Audit - Log the list of flattened dependencies in debug

sverdlov93:
-----

### DIFF
--- a/xray/services/scan.go
+++ b/xray/services/scan.go
@@ -198,11 +198,14 @@ func FlattenGraph(graph []*GraphNode) ([]*GraphNode, error) {
 	for _, node := range graph {
 		populateUniqueDependencies(node, allDependencies)
 	}
-	jsonList, err := json.Marshal(maps.Keys(allDependencies))
-	if err != nil {
-		return nil, errorutils.CheckError(err)
+	if log.GetLogger().GetLogLevel() == log.DEBUG {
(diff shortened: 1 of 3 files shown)


omerzi (approved):

xray/services/scan.go
```diff
@@ -192,12 +193,17 @@ type GraphNode struct {
...
 	allDependencies := map[string]*GraphNode{}
 	for _, node := range graph {
 		populateUniqueDependencies(node, allDependencies)
 	}
-	return []*GraphNode{{Id: "root", Nodes: maps.Values(allDependencies)}}
+	jsonList, err := json.Marshal(maps.Keys(allDependencies))
+	if err != nil {
+		return nil, errorutils.CheckError(err)
+	}
+	log.Debug("Flat dependencies list:\n" + clientutils.IndentJsonArray(jsonList))
```
eyalbe4, line 205: Flat dependencies list: --> Flattened list dependencies to be scanned by Xray:

I also suggest we run the following lines of code only if the log is set to debug, to save ptocessing time. For this, may need to add a new `log.IsDebugEnabled` API - 
```
	jsonList, err := json.Marshal(maps.Keys(allDependencies))
	if err != nil {
		return nil, errorutils.CheckError(err)
	}
```

eyalbe4 (changes requested):

PR #745, opened 2023-04-17
Add Golangci-lint static code analysis and fix found issues

sverdlov93:
-----


utils/io/multifilereader.go
```diff
@@ -1,85 +0,0 @@
-package io
```
Or-Geva, line 1: What is the reason for removing this file?

utils/io/multifilereader_test.go
```diff
@@ -1,67 +0,0 @@
-package io
```
Or-Geva, line 1: What is the reason for removing this file?

utils/tests/utils.go
```diff
@@ -31,7 +31,11 @@ func StartHttpServer(handlers HttpServerHandlers) (int, error) {
 			panic(err)
 		}
 	}()
-	return listener.Addr().(*net.TCPAddr).Port, nil
+	tcpAddr, ok := listener.Addr().(*net.TCPAddr)
+	if !ok {
+		panic("Couldn't assert listener address to tcpAddr")
```
Or-Geva, line 36: We should avoid throwing a panic on the main thread. In contrast to the above panic on line 31, which refers to a subthread.

staticcheck.conf
```diff
@@ -1,2 +0,0 @@
-# Temporary: "-SA1019"
```
Or-Geva, line 1: It was deleted because it was the old static-analysis configuration and we are using a different one now?

Or-Geva (changes requested):

PR #794, opened 2023-07-02
Edit prepare exclusions pattern function

sarao1310:
-----

### DIFF
--- a/artifactory/services/fspatterns/utils.go
+++ b/artifactory/services/fspatterns/utils.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"path/filepath"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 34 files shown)


eyalbe4 (changes requested):

artifactory/services/fspatterns/utils.go
```diff
@@ -30,14 +29,14 @@ func ListFiles(rootPath string, isRecursive, includeDirs, isSymlink bool, exclud
 	return filterFiles(paths, excludePathPattern)
 }
 
-// Transform to regexp and prepare Exclude patterns to be used
-func PrepareExcludePathPattern(params serviceutils.FileGetter) string {
+// Transform to regexp and prepare Exclude patterns to be used, exclusion patterns must be absolute paths.
+func PrepareExcludePathPattern(exclusions []string, patternType utils.PatternType, isRecursive bool) (string, error) {
```
eyalbe4, line 33: isRecursive --> recursive

This function never returned an error, so it probably shouldn't declare it.

This PR seems to be changing nothing except for this function's prototype change. Are you sure this PR is needed? Please reply here.

eyalbe4 (changes requested):

artifactory/services/fspatterns/utils.go
sarao1310: yes, We now use PrepareExcludePathPattern in the ArchiveProject function in cli-core (that has been called by the GoPublish command), and GoPublishParams does not inherit from FileGetter, so I change the prototype to pass only parameters we really use in PrepareExcludePathPattern.

I'll change the return value.

PR #815, opened 2023-08-24
Improve Audit and Scan performance for large projects

sverdlov93:
-----


utils/tests/utils.go
```diff
@@ -139,10 +139,10 @@ func ChangeDirAndAssert(t *testing.T, dirPath string) {
...
 
 // ChangeDirWithCallback changes working directory to the given path and return function that change working directory back to the original path.
-func ChangeDirWithCallback(t *testing.T, wd, dirPath string) func() {
-	ChangeDirAndAssert(t, dirPath)
+func ChangeDirWithCallback(t *testing.T, callbackDir, chdirPath string) func() {
+	ChangeDirAndAssert(t, chdirPath)
 	return func() {
-		ChangeDirAndAssert(t, wd)
+		ChangeDirAndAssert(t, callbackDir)
 	}
```
omerzi, line 146: IMO, `origin`/`source`, `destination` sounds clearer.

xray/services/scan.go
```diff
@@ -165,50 +169,12 @@ type XrayGraphScanParams struct {
 	ProjectKey             string
 	Watches                []string
 	ScanType               ScanType
-	Graph                  *xrayUtils.GraphNode
+	AuditGraph             *xrayUtils.GraphNode
+	BinaryGraph            *xrayUtils.BinaryGraphNode
```
omerzi, line 173: I understand that you want to provide a clarification where one graph is used in the Audit command, and the second one refers to Binaries. So, what do you think about changing the names to: DependenciesGraph and BinariesGraph? In my opinion, the name "AuditGraph" might not be the most appropriate choice, as it might not directly convey that this graph specifically contains dependencies. What do you think?

PR #883, opened 2023-12-27
XRAY-36857 - Add JF_MSI env var

orz25:
-----


xray/services/scan.go
```diff
@@ -111,6 +112,10 @@ func (ss *ScanService) ScanGraph(scanParams XrayGraphScanParams) (string, error)
 			return "", fmt.Errorf("failed sending Git Info to XSC service, error: %s ", err.Error())
 		}
 		scanParams.MultiScanId = multiScanId
+		err = os.Setenv("JF_MSI", multiScanId)
+		if err != nil {
+			return "", fmt.Errorf("failed setting msi as environment variable")
+		}
```
eyalbe4, line 118: The above code can be simplified as follows. In addition, the error created should be wrapped with `errorutils.CheckErrorf(err)` iin this case, beucase `os.Setenv`, which may return the error, is an external API to the `jfrog-client-go` codebase. Also notice that in the above code the original error is lost, and the below code fixes that as well.
```
if err = os.Setenv("JF_MSI", multiScanId); err != nil {
	errorutils.CheckErrorf("failed setting msi as environment variable. Cause: %s", err.Error())
}
```

eyalbe4 (changes requested):
Before we can merge this PR, you should go ahead and create two additional PRs for the *jfrog-cli-core* and *jfrog-cli* projects, so that we can see that the tests that those projects include, and may be impacted by this change, as passing.
You'll need to use the `replace` statements in the `go.mod` files of those projects, to reference this PR's branch (your fork) for both of your `jfrog-client-go` and `jfrog-cli-core`. It'll be helpfull if you link all of the 3 PRs by adding referencing them in the descriptions of the PRs.

xray/services/scan.go
```diff
@@ -113,7 +113,7 @@ func (ss *ScanService) ScanGraph(scanParams XrayGraphScanParams) (string, error)
 		}
 		scanParams.MultiScanId = multiScanId
 		if err = os.Setenv("JF_MSI", multiScanId); err != nil {
-			errorutils.CheckErrorf("failed setting msi as environment variable. Cause: %s", err.Error())
+			log.Warn(errorutils.CheckErrorf("failed setting msi as environment variable. Cause: %s", err.Error()))
```
eyalbe4, line 116: We try to reduce the usage of `log.Error` and `log.Warn` to the minimum.
I suggest changing the above line to - 
```
 return errorutils.CheckErrorf("failed setting msi as environment variable. Cause: %s", err.Error())
```
You probably wouldn't like to have code flow continue after such an error.

eyalbe4 (changes requested):

xray/services/scan.go
hadarshjfrog: Suggestion: If we'll do `log.Debug` - would it be a good solution?
We wouldn't want to drop it - but do have the possibility to debug when necessary

orz25: I changed the log to Debug, since we don't want to stop the whole scan.

xray/services/scan.go
```diff
@@ -111,6 +112,9 @@ func (ss *ScanService) ScanGraph(scanParams XrayGraphScanParams) (string, error)
 			return "", fmt.Errorf("failed sending Git Info to XSC service, error: %s ", err.Error())
 		}
 		scanParams.MultiScanId = multiScanId
+		if err = os.Setenv("JF_MSI", multiScanId); err != nil {
+			log.Debug(errorutils.CheckErrorf("failed setting msi as environment variable. Cause: %s", err.Error()))
```
attiasas, line 116:
```suggestion
			log.Debug(errorutils.CheckErrorf("failed setting MSI as environment variable. Cause: %s", err.Error()))
```

xray/services/scan.go
```diff
@@ -111,6 +112,9 @@ func (ss *ScanService) ScanGraph(scanParams XrayGraphScanParams) (string, error)
 			return "", fmt.Errorf("failed sending Git Info to XSC service, error: %s ", err.Error())
 		}
 		scanParams.MultiScanId = multiScanId
```
attiasas, line 114: is it OK to pass it as param and environment variable as well?

orz25: Removed all places we use multi_scan_id param

Issue #707, opened 2023-02-16
Improve SetUrl Logic to handle `$URI` and `$URI`.


joshrosso:
**Is your feature request related to a problem? Please describe.**

As a new user, I struggled to determine why I was getting `404`s on my API requests. Turns out, I was setting the URL as:

```go
ac := auth.NewArtifactoryDetails()
ac.SetUrl("https://arctir.jfrog.io/artifactory")
```

And thus the config was resolving as `https://${ORG_NAME}.jfrog.io/artifactoryapi`.

I'd expect [SetUrl](https://github.com/jfrog/jfrog-client-go/blob/ac964bad46f2549ac8d1b50c2b0e3b632cba0f45/auth/servicedetails.go#L129-L131) to handle both of these cases:

https://github.com/jfrog/jfrog-client-go/blob/ac964bad46f2549ac8d1b50c2b0e3b632cba0f45/auth/servicedetails.go#L129-L131

Just figured I'd pass this feedback/suggestion on from the perspective of a new user 😃. If it's not desirable behavior for the project, feel free to close. 

Thanks for your work on maintaining this library!

AF250329:
yes... it took me hours to understand why HTTP request fails.... all because of the missing / at the end

sverdlov93:
Hi @joshrosso @AF250329 
Thanks for reporting this issue.
Our apologies for the inconvenience caused by this.
This [PR](https://github.com/jfrog/jfrog-client-go/pull/887) should fix it.
I will update this thread once the next release will be out.

sverdlov93:
Hi @joshrosso @AF250329
Version 1.36.0 has been released!
We are looking forward to your feedback on it.

PR #887, opened 2024-01-14
Add trailing slash if needed on SetUrl method

sverdlov93:
-----
following https://github.com/jfrog/jfrog-client-go/issues/707

Also:
1. Remove the leading slash if it exists on the target path to avoid wrong target parsing.
2. Comments have been added to clarify some potentially confusing lines of code.
3. Fix failing terraform update test.


artifactory/services/utils/specutils.go
```diff
@@ -91,7 +91,9 @@ func (params *CommonParams) SetPattern(pattern string) {
 }
 
 func (params *CommonParams) SetTarget(target string) {
-	params.Target = target
+	// Target Specifies the target path in Artifactory in the following format: <repository name>/<repository path>, so it cannot start with a slash.
+	// Remove leading slash if exists
+	params.Target = strings.TrimPrefix(target, "/")
```
yahavi, line 96: The Target is a public field and can be set directly.
Let's ensure that the code works without this change.

Same comment for the SetUrl in CommonConfigFields.

yahavi (changes requested):

PR #917, opened 2024-02-29
Artifactory Release Lifecycle Management  - Add Export bundle function

EyalDelarea:
-----

## Added Export Release Lifecycle Export function

- `jf rbe <bundle_name>_<bundle_version> [target pattern ]` will download the archive

- The process is triggering the export process ( which can take a while ) and once ready, download the archive.


lifecycle/services/export.go
```diff
@@ -0,0 +1,174 @@
...
+	ReleaseBundleExportParams
+}
+
+type exportStatusOperation struct {
+	reqBody     RbExportBody
+	queryParams CommonOptionalQueryParams
+}
+
+func (exs *exportStatusOperation) getOperationRestApi() string {
+	fullUrl, err := url.JoinPath(distributionBaseApi, exportReleaseBundleStatusEndpoint, exs.reqBody.ReleaseBundleName, exs.reqBody.ReleaseBundleVersion)
```
RobiNino, line 48: Opt to use `path.Join()` here instead, which will also remove the need to handle an error.
As a side note - our convention is to never panic, rather always return the error.

lifecycle/services/export.go
```diff
@@ -0,0 +1,174 @@
...
+	return ""
+}
+
+type exportOperation struct {
+	reqBody     RbExportBody
+	queryParams CommonOptionalQueryParams
+}
+
+func (exp *exportOperation) getOperationRestApi() string {
+	fullUrl, err := url.JoinPath(distributionBaseApi, exportReleaseBundleEndpoint, exp.reqBody.ReleaseBundleName, exp.reqBody.ReleaseBundleVersion)
```
RobiNino, line 77: Same comment as above

lifecycle/services/operation.go
```diff
@@ -35,7 +36,13 @@ type ReleaseBundleOperation interface {
 	getSigningKeyName() string
 }
 
-func (rbs *ReleaseBundlesService) doOperation(operation ReleaseBundleOperation) ([]byte, error) {
+type operationParams struct {
+	httpMethod  string
+	contentType string
+}
+
+func (rbs *ReleaseBundlesService) doOperation(operation ReleaseBundleOperation, operationParams ...operationParams) (body []byte, err error) {
```
RobiNino, line 44: The intention here to reuse code is great, but I think it will be simpler and more self-explanatory if we create two functions, each for every http method. Both can use a third function with a common code.

lifecycle/services/export.go
```diff
@@ -0,0 +1,174 @@
...
+	operationExtraParams := operationParams{
+		httpMethod:  http.MethodGet,
+		contentType: "application/json",
+	}
+	log.Debug("Getting Release Bundle Export status...")
+	respBody, err := rbs.doOperation(operation, operationExtraParams)
+	if err != nil {
+		return
+	}
+	err = json.Unmarshal(respBody, &exportedStatusResponse)
```
RobiNino, line 172:
```suggestion
	err = errorutils.CheckError(json.Unmarshal(respBody, &exportedStatusResponse))
```
Cover potential errors where they first appear in our code (generated by us / third party packages) with the `CheckError` function. This function is a debug tool that can be used to identify where an error was thrown from.

lifecycle/manager.go
```diff
@@ -107,3 +107,8 @@ func (lcs *LifecycleServicesManager) RemoteDeleteReleaseBundle(params distributi
 	rbService := lifecycle.NewReleaseBundlesService(lcs.config.GetServiceDetails(), lcs.client)
 	return rbService.RemoteDeleteReleaseBundle(params, dryRun)
 }
+
+func (lcs *LifecycleServicesManager) ExportReleaseBundle(releaseBundleExportParams *lifecycle.ReleaseBundleExportParams, queryParams lifecycle.CommonOptionalQueryParams) (exportResponse lifecycle.ReleaseBundleExportedStatusResponse, err error) {
```
RobiNino, line 111: I believe it will be best to follow the convention where the release bundle details are first in the service signature, separated from other params.
In addition, I suggest not use a pointer to the params if we don't need the reference. Rather to pass by value.

```suggestion
func (lcs *LifecycleServicesManager) ExportReleaseBundle(rbDetails lifecycle.ReleaseBundleDetails, releaseBundleExportParams lifecycle.ReleaseBundleExportParams, queryParams lifecycle.CommonOptionalQueryParams) (exportResponse lifecycle.ReleaseBundleExportedStatusResponse, err error) {
```

lifecycle/services/export.go
```diff
@@ -0,0 +1,174 @@
...
+	DownloadUrl string           `json:"download_url,omitempty"`
+}
+
+type ReleaseBundleExportParams struct {
+	ReleaseBundleDetails
+	Modifications `json:"modifications"`
+}
+
+type RbExportBody struct {
+	ReleaseBundleExportParams
```
RobiNino, line 39: If I'm not mistaken, the release bundle details will be sent here although they are not expected to be part of the body.  Better to send only what is expected.

lifecycle/services/export.go
```diff
@@ -0,0 +1,174 @@
...
+func (exp *exportOperation) getSigningKeyName() string {
+	return ""
+}
+
+func (rbs *ReleaseBundlesService) ExportReleaseBundle(rlExportParams *ReleaseBundleExportParams, queryParams CommonOptionalQueryParams) (exportResponse ReleaseBundleExportedStatusResponse, err error) {
+	// Check the current status
+	if exportResponse, err = rbs.getExportedReleaseBundleStatus(rlExportParams, queryParams); err != nil {
+		return
+	}
+	// Trigger export if needed
```
RobiNino, line 105: We can simplify the code a bit by checking here.
```
if exportResponse.Status == ExportCompleted {
    return
}
```
Then the last if is not necessary.

lifecycle/services/export.go
```diff
@@ -0,0 +1,174 @@
...
+	// Wait for export to finish
+	if exportResponse.Status != ExportCompleted {
+		if exportResponse, err = rbs.checkExportedStatusWithRetries(rlExportParams, queryParams); err != nil {
+			return
+		}
+	}
+	return
+}
+
+func (rbs *ReleaseBundlesService) checkExportedStatusWithRetries(rlExportParams *ReleaseBundleExportParams, queryParams CommonOptionalQueryParams) (response ReleaseBundleExportedStatusResponse, err error) {
```
RobiNino, line 120: Consider renaming to `waitForExport` (retries usually indicates something went wrong with the request itself)

RobiNino (changes requested):
Looks good! See my inline comments and please add the new service to README.md

RobiNino (changes requested):

RobiNino (changes requested):

PR #921, opened 2024-03-11
Artifactory Release Lifecycle Management - Add Import bundle function

EyalDelarea:
-----

Added import release bundle archive functionality


artifactory/services/release.go
```diff
@@ -0,0 +1,84 @@
...
+	importRestApiEndpoint = "api/release/import"
+	// TODO change this when the bug is fixed
+	// https://jfrog-int.atlassian.net/browse/JR-8542
+	tempImportRestApiEndpoint = "ui/api/v1/ui/release/import"
+	octetStream               = "application/octet-stream"
+)
+
+type ReleaseService struct {
+	client     *jfroghttpclient.JfrogHttpClient
+	ArtDetails auth.ServiceDetails
```
RobiNino, line 27: Can be made private

artifactory/emptymanager.go
```diff
@@ -105,6 +105,7 @@ type ArtifactoryServicesManager interface {
 	FileList(relativePath string, optionalParams utils.FileListParams) (*utils.FileListResponse, error)
 	GetStorageInfo() (*utils.StorageInfo, error)
 	CalculateStorageInfo() error
+	ReleaseBundleImport(string) error
```
RobiNino, line 108: I suggest to rename to `ImportReleaseBundle` to follow the convention

artifactory/services/release.go
```diff
@@ -0,0 +1,84 @@
+package services
```
RobiNino, line 1: I suggest renaming the file to `releasebundle`

artifactory/services/release.go
```diff
@@ -0,0 +1,84 @@
...
+	return &ReleaseService{client: client, ArtDetails: artDetails}
+}
+
+func (rs *ReleaseService) GetJfrogHttpClient() *jfroghttpclient.JfrogHttpClient {
+	return rs.client
+}
+
+func (rs *ReleaseService) ImportReleaseBundle(filePath string) (err error) {
+	// Load desired file
+	content, err := fileutils.ReadFile(filePath)
```
RobiNino, line 50: Just to make sure - did you happen to check if the file name of the archive can be changed or is expected to be left unchanged?

RobiNino (changes requested):
Looks good! Lets add a simple unit test and mark this PR as ready so the tests will run

artifactory/services/release.go
```diff
@@ -0,0 +1,84 @@
...
+		return
+	}
+	// When a release bundle already exists, don't return an error message of failure.
+	if resp.StatusCode == http.StatusBadRequest {
+		response := ErrorResponseWithMessage{}
+		if err = json.Unmarshal(body, &response); err != nil {
+			return
+		}
+		if response.Errors[0].Message == conflictErrorMessage {
+			log.Warn("Bundle already exists, did not upload a new bundle")
```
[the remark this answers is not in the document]
EyalDelarea: This is an answer to unexpected API response, as we are getting 400 when there is already a bundle. instead of 409 which can be handled as conflict.

So i was thinking of checking that the 400 status code is indeed a conflict.
And if that's the case - ignore it.
Otherwise it will be an error and logged in the error trace.

artifactory/services/release.go
EyalDelarea: When you import a bundle you can't change it's name.
And even if you do, if the metadata inside the bundle doesn't match it fails to import. 👍

EyalDelarea: I've added a unit test for this case.

artifactory/services/release.go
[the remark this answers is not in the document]
RobiNino: OK, understood!

PR #927, opened 2024-03-25
Fix set nil for 'checkRedirect' when no value exists, tmp fix.

asafambar:
- [X] All [static analysis checks](https://github.com/jfrog/jfrog-client-go/actions/workflows/analysis.yml) passed.
-----

### DIFF
--- a/http/httpclient/client.go
+++ b/http/httpclient/client.go
@@ -180,8 +180,9 @@ func (jc *HttpClient) doRequest(req *http.Request, content []byte, followRedirec
 	}
 
 	resp, err = jc.client.Do(req)
-	jc.client.CheckRedirect = nil
-
+	if jc.client.CheckRedirect != nil {
+		jc.client.CheckRedirect = nil
+	}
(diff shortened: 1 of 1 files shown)


asafambar:
This PR is a temp solution for data race found by go tool, it prevent the tool from failing.
in curation we have many requests in parallel using the same http client, tone of the struct values 'CheckRedirect' is set to nil after each request, as for the usage of the use today with the client, the value is always nil, so there is no need to write any nil value again which cause data race,
For better future use of this client to have parallel requests safely (only on post requests or specific config use it) ,  this should not be changed after each requests but be determined on the Initialization of the client.

sverdlov93:
@asafambar fixed it - https://github.com/jfrog/jfrog-client-go/commit/ca94b4567e7b5ff200b46d76f27a03e990eeb3b8#diff-98b36a502daf580571eea44435a154850e1ee83f21d7c2b56d9bf976f7b4adc2R185

PR #937, opened 2024-04-11
Support attaching Uber Trace ID to every request

RobiNino:
-----

### DIFF
--- a/http/httpclient/client.go
+++ b/http/httpclient/client.go
@@ -36,7 +36,13 @@ type HttpClient struct {
 const (
 	apiKeyPrefix        = "AKCp8"
 	apiKeyMinimalLength = 73
+	uberTraceIdHeader   = "uber-trace-id"
 )
+
+// If set, the Trace ID token will be used to generate an Uber Trace ID header which will be attached to every request.
+// This allows users to easily identify which logs on the server side are related to requests sent from this client.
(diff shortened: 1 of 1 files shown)


http/httpclient/client.go
```diff
@@ -36,8 +36,14 @@ type HttpClient struct {
 const (
 	apiKeyPrefix        = "AKCp8"
 	apiKeyMinimalLength = 73
+	uberTraceIdHeader   = "uber-trace-id"
 )
 
+// If set, the Trace ID token will be used to generate an Uber Trace ID header which will be attached to every request.
+// This allows users to easily identify which logs on the server side are related to requests sent from this client.
+// The expected token value is 16 hexadecimal characters.
+var TraceIdToken string
```
yahavi, line 45:
```suggestion
var UberTraceIdToken string
```
+
Let's add a function:
```go
func SetUberTraceIdToken(traceIdToken string)
```
This function will set the uberTraceIdToken to be:
```go
uberTraceIdToken = fmt.Sprintf("%s:%s:0:0", traceIdToken, traceIdToken))
```

By that, we will not have to recalculate the string for every HTTP request.

yahavi (changes requested):

yahavi (approved):
LGTM

PR #951, opened 2024-05-09
Fix username extraction on OIDC tokens with a group scope

yahavi:
-----
Fix https://github.com/jfrog/setup-jfrog-cli/issues/146.

Previously, we assumed all JWT tokens followed this **subject** structure:
```
jfac@<some-characters>/users/<user-name>
```
However, OIDC tokens with group scopes might follow this **subject** structure instead:
```
<provider-name>/<identity-mapping>
```
This structure can serve as a standard username.

### DIFF
--- a/auth/authutils_test.go
+++ b/auth/authutils_test.go
@@ -33,6 +33,7 @@ func TestExtractUsernameFromAccessToken(t *testing.T) {
 		{"tempuser", token4, false},
 		{"yahav/test-repo", token5, false},
 	}
+	// Discard output logging to prevent negative logs
 	previousLog := tests.RedirectLogOutputToNil()
 	defer func() {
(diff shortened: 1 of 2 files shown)


auth/authutils_test.go
```diff
@@ -27,7 +31,12 @@ func TestExtractUsernameFromAccessToken(t *testing.T) {
 		{"", token2, true},
 		{"", token3, true},
 		{"tempuser", token4, false},
+		{"yahav/test-repo", token5, false},
 	}
+	previousLog := tests.RedirectLogOutputToNil()
```
eyalbe4, line 36: Let's add a comment explaining why we're redirecting the log output.

eyalbe4 (changes requested):

PR #964, opened 2024-06-24
Fix http client redirect race condition

sverdlov93:
-----

The HTTP client is a shared resource between go routines, so to handle this override we clone it.


http/httpclient/client.go
```diff
@@ -179,16 +179,22 @@ func (jc *HttpClient) doRequest(req *http.Request, content []byte, followRedirec
...
 	if !followRedirect || (followRedirect && req.Method == http.MethodPost) {
-		jc.client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
-			redirectUrl = req.URL.String()
-			return errors.New("redirect")
+		// The jc.client is a shared resource between go routines, so to handle this override we clone it.
+		client = &http.Client{
+			Transport: jc.client.Transport,
+			Timeout:   jc.client.Timeout,
+			Jar:       jc.client.Jar,
+			CheckRedirect: func(req *http.Request, via []*http.Request) error {
```
yahavi, line 190: Let's use a new method cloneHttpClient for better readability 
```suggestion
		client = cloneHttpClient(jc.client)
		client.CheckRedirect: func(req *http.Request, via []*http.Request) error {
			redirectUrl = req.URL.String()
			return errors.New("redirect")
		}
```

yahavi (changes requested):

PR #996, opened 2024-08-06
Unite http content type header code

sverdlov93:
-----
Converted AddHeader to a method and united content type json usages:
image


artifactory/services/props.go
```diff
@@ -44,21 +44,11 @@ func (ps *PropsService) GetThreads() int {
 }
 
 func (ps *PropsService) SetProps(propsParams PropsParams) (int, error) {
-	log.Info("Setting properties...")
```
RobiNino, line 47: I think all the logs in this file are useful, why don't you like them?

RobiNino (changes requested):

artifactory/services/props.go
sverdlov93: setProps calls put/delete props and they already have a log of their own:

- log.Info(logMsgPrefix+"Deleting properties on:", relativePath)
- log.Info(logMsgPrefix+"Setting properties on:", relativePath)

@RobiNino

PR #1005, opened 2024-08-15
Improve token and build info service errors

sverdlov93:
-----


tests/artifactoryctx_test.go
```diff
@@ -40,8 +39,6 @@ func testCtxTimeout(t *testing.T) {
 	assert.NoError(t, err)
 	time.Sleep(time.Millisecond * 300)
 	_, err = sm.GetVersion()
-	assert.Error(t, err)
-	if !errors.Is(err, context.DeadlineExceeded) {
-		t.Fail()
-	}
+	// Expect timeout error
+	assert.ErrorContains(t, err, context.DeadlineExceeded.Error())
```
eyalbe4, line 43: Are you sure this change is safe?
`errors.Is` os very different than `ErrorContains`.
The former checks the type of the error, and the later checks a string inside the error message.

eyalbe4 (changes requested):

tests/artifactoryctx_test.go
sverdlov93: yeah because the new error now is wrapped by a more informative one
@eyalbe4

eyalbe4: :+1:

PR #1006, opened 2024-08-16
Add and use a new IsPathAccessible API before accessing the .git dir

eyalbe4:
This is an attempt to fix https://github.com/jfrog/jfrog-cli/issues/2442.
The PR adds a new `IsPathAccessible` API, in addition to the existing `IsPathExists` functions. This is because the existing `IsPathExists` returns `false` in case of any error that occurred while accessing the directory, and not only in the case that the directory doesn't exist.
The new `IsPathAccessible` returns `false` in case the directory exists, but the user doesn't have permissions to access it. The usage of the new API by the Artifactory Upload code flow is expected to resolve the above issue.

### DIFF
--- a/utils/io/fileutils/files_test.go
+++ b/utils/io/fileutils/files_test.go
@@ -15,10 +15,16 @@ import (
 )
 
 func TestIsPathExistsAndIsPathAccessible(t *testing.T) {
+	var symlinkPath string
+	symlinkCreated := false
+
 	// Create a temporary file
 	tempFile, err := os.CreateTemp("", "testfile")
 	assert.NoError(t, err)
 	defer func() {
+		if symlinkCreated {
+			assert.NoError(t, os.Remove(symlinkPath))
+		}
 		assert.NoError(t, os.Remove(tempFile.Name()))
 	}()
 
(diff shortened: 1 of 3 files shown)


utils/io/fileutils/files_test.go
```diff
@@ -1,18 +1,74 @@
...
+
+	// Test for a non-existing file
+	assert.False(t, IsPathExists(tempFile.Name()+"_nonexistent", false))
+
+	// Create a temporary directory
+	tempDir, err := os.MkdirTemp("", "testdir")
+	assert.NoError(t, err)
+	defer func() {
+		assert.NoError(t, os.RemoveAll(tempDir))
+	}()
```
sverdlov93, line 48: t.TempDir() is nicer and its automatically handles cleanup after the test.

utils/io/fileutils/files_test.go
```diff
@@ -1,18 +1,74 @@
...
 	"strings"
 	"testing"
 
+	biutils "github.com/jfrog/build-info-go/utils"
+	"github.com/jfrog/jfrog-client-go/utils/io"
+
 	"github.com/stretchr/testify/assert"
 )
 
+func TestIsPathExistsAndIsPathAccessible(t *testing.T) {
```
sverdlov93, line 17: for isAccessible should we create a file without permissions and test that it returns false as expected?

eyalbe4: I tried adding a test that creates a file without permissions, but you have no idea how challenging it is to test this. The major challenge comes from the fact that Windows and Linux/MacOS treat files without permissions completely differently. It is hard to make all OS work inside a single test. The only option I found was to split the test into two files, so that only one of them compilers for Windows. This is due to a package that can handle Windows locking. I ended up giving up and just testing the IsPathAccessible API with non existing files. It's just not worth the complexity @sverdlov93.

PR #1019, opened 2024-09-11
Handle archive upload general progress bar counter incrementation

sverdlov93:
-----
Today, all files uploaded as part of an archive are counted for the total number of tasks (denominator), but the numerator are counted only at the end of the zip upload as 1 task.
So for 60k files it's 1/60000 till the end of the upload.
This PR fixes it.


artifactory/services/upload.go
```diff
@@ -993,6 +993,10 @@ func (us *UploadService) addFileToZip(artifact *clientutils.Artifact, progressPr
 	if errorutils.CheckError(err) != nil {
 		return
 	}
+	// Increment general progress by 1 for each file added to the zip.
+	if us.Progress != nil {
```
yahavi, line 997: `CreateUploadAsZipFunc` calls `addFileToZip` twice. Wouldn't this increment the progress twice? Or is that intentional?

sverdlov93: check now @yahavi

artifactory/services/upload.go
```diff
@@ -886,12 +889,17 @@ func (us *UploadService) CreateUploadAsZipFunc(uploadResult *utils.Result, targe
 
 // Reads files and streams them as a ZIP to a Reader.
 // archiveDataReader is a ContentReader of UploadData items containing the details of the files to stream.
+// isZipDryRun - If true, this function is run as part of the dry-run process, to
 // saveFilesPathsFunc (optional) is a func that is called for each file that is written into the ZIP, and gets the file's local path as a parameter.
-func (us *UploadService) readFilesAsZip(archiveDataReader *content.ContentReader, progressPrefix string, flat, symlink bool,
+func (us *UploadService) readFilesAsZip(archiveDataReader *content.ContentReader, isZipDryRun, flat, symlink bool,
```
eyalbe4, line 894: isZipDryRun --> dryRun

In addition to removing the redundant "zip" section from the argument name, that is unnecessary beucase the function's content is already related to the zip, it's preferred to avoid adding the "is" prefix to boolean variables.

Let's change the argument's description as well.

eyalbe4 (changes requested):

artifactory/services/upload.go
sverdlov93: @eyalbe4 its on purpose. We already have a dryRun expression among JFrog CLI command so I didn't want to confuse with that known flag.
Do you have another idea?

PR #1036, opened 2024-10-31
Remove remaining temp file after Artifactory download

eyalbe4:
Fix a bug where after a file is downloaded from Artifactory, there's one temporary file that remains in the OS temp dir. This fix removes this temporary file.

### DIFF
--- a/artifactory/services/download.go
+++ b/artifactory/services/download.go
@@ -3,16 +3,17 @@ package services
 import (
 	"errors"
 	"fmt"
-	"github.com/jfrog/build-info-go/entities"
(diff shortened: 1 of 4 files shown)


Or-Geva (changes requested):
The fix looks good, @eyalbe4. However, I think we should add a test for this flow. The test could be included here or, if more convenient, in JFrog CLI repository. It should verify that no temporary files remain once the download is completed.

eyalbe4:
Since implementing such a test is very challenging, @Or-Geva and I decided not to implement it.

PR #1046, opened 2024-11-14
Transition Xsc to Xray

attiasas:
-----

* Remove the "404" error when `XSC` is not avaliable
* Remove old unused API
* Add the new API URL's to the XSC service in its new location as an inner service in Xray


xsc/services/utils/utils.go
```diff
@@ -0,0 +1,34 @@
...
+
+const (
+	XraySuffix                        = "/xray/"
+	xscSuffix                         = "/xsc/"
+	apiV1Suffix                       = "api/v1"
+	XscInXraySuffix                   = apiV1Suffix + xscSuffix
+	MinXrayVersionXscTransitionToXray = "3.108.0"
+)
+
+// From Xray version 3.108.0, XSC is transitioning to Xray as inner service. This function will return the backward compatibility URL.
```
orz25, line 19: I think the last part of the comment should be :
This function will return compatible URL

xray/services/xsc/xsc.go
```diff
@@ -0,0 +1,52 @@
+package xsc
```
orz25, line 1: I think we should add a comment explaining the difference between XscInnerService and XscServicesManager. It's a bit confusing

xray/services/scan.go
```diff
@@ -186,42 +180,19 @@ func (ss *ScanService) GetScanGraphResults(scanId string, includeVulnerabilities
...
-	xscResponse := XscPostContextResponse{}
-	if err = json.Unmarshal(body, &xscResponse); err != nil {
-		return "", errorutils.CheckError(err)
-	}
-	return xscResponse.MultiScanId, err
-}
-
 // IsXscEnabled will try to get XSC version. If route is not available, user is not entitled for XSC.
-func (ss *ScanService) IsXscEnabled() (xsxVersion string, err error) {
+func (ss *ScanService) IsXscEnabled() (xscVersion string, err error) {
```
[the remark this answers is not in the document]
attiasas: it calls the version API, if we call it anywhere we will do redundant API calls.
I think we don't use it anymore, I will delete it as well.

xray/services/xsc/xsc.go
attiasas: added comment

xsc/services/utils/utils.go
attiasas: done

PR #1061, opened 2024-12-22
Internal API - Visibility System fix & improvements

eyalbe4:
* Fix the API endpoint.
* Improve the API to accept a struct instead of a JSON

### DIFF
--- a/jfconnect/manager.go
+++ b/jfconnect/manager.go
@@ -7,7 +7,7 @@ import (
 )
 
 type Manager interface {
-	PostMetric([]byte) error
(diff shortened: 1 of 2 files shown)


jfconnect/manager.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 type Manager interface {
-	PostMetric([]byte) error
+	PostMetric(services.VisibilityMetric) error
```
sverdlov93, line 10: maybe:
```suggestion
	PostVisibilityMetric(services.VisibilityMetric) error
```

PR #1063, opened 2024-12-25
Internal - Remove the Call Home version validation

eyalbe4:
* Remove the minimum Artifactory version validation before sending the Call Home requests, since the validation moves to the caller of the API. This is done to avoid duplicate "Get Artifactory Version" requests.
* Rename the API functions, to force the callers to adapt to this change.

### DIFF
--- a/artifactory/usage/call_home.go
+++ b/artifactory/usage/call_home.go
@@ -11,8 +11,6 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/io/httputils"
 )
 
-const minArtifactoryVersion = "6.9.0"
-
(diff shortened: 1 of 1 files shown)


EyalDelarea (changes requested):
@eyalbe4,
What happens if, as a future implementer, I fail to check the minimum version? How will I know?
I know the functions names has changed, but it does not enforce a version check.

Also, how will I know the minimum version needed now that it has moved to the core? assuming i'm a developer who do want to validate on my end.

eyalbe4:
@EyalDelarea,
This API is internal, and should not be used independently.

PR #1083, opened 2025-02-13
Support for Build info overwrite for a given build name and number

bhanurp:
-----
Created DeleteBuildInfo where it takes name, number and number of build number Occurrences to delete the build info. The existing API's deletes only 1 build even though there are multiple build info with same name number. numberOfOccurrences param is used to delete n number of such occurrences


artifactory/services/buildinfo.go
```diff
@@ -87,3 +95,37 @@ func (bis *BuildInfoService) PublishBuildInfo(build *buildinfo.BuildInfo, projec
...
+		return err
+	}
+	if err = errorutils.CheckResponseStatusWithBody(resp, body, http.StatusOK, http.StatusCreated, http.StatusNoContent); err != nil {
+		return err
+	}
+	log.Debug("Artifactory response", resp.Status)
+	return nil
+}
+
+func CreateDeleteBuildInfoBody(build *buildinfo.BuildInfo, projectKey string, buildNumberFrequency int) DeleteBuildInfoBody {
```
RobiNino, line 119: This function can be made private.

artifactory/services/buildinfo.go
```diff
@@ -87,3 +95,37 @@ func (bis *BuildInfoService) PublishBuildInfo(build *buildinfo.BuildInfo, projec
...
+	buildNumbers := make([]string, 0, buildNumberFrequency)
+	for i := 0; i < buildNumberFrequency; i++ {
```
RobiNino, line 121: I can't seem to understand what `buildNumberFrequency` does from the code, and there is no documentation on it. Please add some comments and documentation in the README regarding this input. We can also first discuss the functionality it adds over the REST API if you prefer.

artifactory/services/buildinfo.go
```diff
@@ -87,3 +95,37 @@ func (bis *BuildInfoService) PublishBuildInfo(build *buildinfo.BuildInfo, projec
 	log.Debug("Artifactory response:", resp.Status)
 	return summary, nil
 }
+
+func (bis *BuildInfoService) DeleteBuildInfo(build *buildinfo.BuildInfo, projectKey string, buildNumberFrequency int) error {
+	params := CreateDeleteBuildInfoBody(build, projectKey, buildNumberFrequency)
+	content, err := json.Marshal(params)
+	if err != nil {
+		return err
```
RobiNino, line 103:
```suggestion
		return errorutils.CheckError(err)
```
We wrap error where they originally returned by this function. It serves a debugging functionality, which we can go over in our next session.

artifactory/services/buildinfo.go
```diff
@@ -87,3 +95,37 @@ func (bis *BuildInfoService) PublishBuildInfo(build *buildinfo.BuildInfo, projec
...
+	}
+	httpClientsDetails := bis.GetArtifactoryDetails().CreateHttpClientDetails()
+	utils.SetContentType("application/json", &httpClientsDetails.Headers)
```
RobiNino, line 106: Using `httpClientsDetails.SetContentTypeApplicationJson()` maybe cleaner

artifactory/services/buildinfo.go
```diff
@@ -87,3 +95,37 @@ func (bis *BuildInfoService) PublishBuildInfo(build *buildinfo.BuildInfo, projec
...
+	resp, body, err := bis.client.SendPost(bis.GetArtifactoryDetails().GetUrl()+"api/build/delete", content, &httpClientsDetails)
```
RobiNino, line 107: Since this service allows setting `DryRun`, we might want to support it by logging the content and returning before sending the request (similar to publish build info)

artifactory/services/buildinfo.go
```diff
@@ -87,3 +95,37 @@ func (bis *BuildInfoService) PublishBuildInfo(build *buildinfo.BuildInfo, projec
...
+	if err != nil {
+		log.Error("Error occurred while deleting build info", err)
```
RobiNino, line 109: I suggest to return a new error with this prefix and the original error message instead of logging (which will display multiple error messages in the log)

artifactory/services/buildinfo.go
```diff
@@ -87,3 +95,37 @@ func (bis *BuildInfoService) PublishBuildInfo(build *buildinfo.BuildInfo, projec
...
+		return err
+	}
+	if err = errorutils.CheckResponseStatusWithBody(resp, body, http.StatusOK, http.StatusCreated, http.StatusNoContent); err != nil {
```
RobiNino, line 112:
```suggestion
	if err = errorutils.CheckResponseStatusWithBody(resp, body, http.StatusOK,  http.StatusNoContent); err != nil {
```

RobiNino (changes requested):

artifactory/services/buildinfo.go
bhanurp: we have seen this but never understood its purpose properly

RobiNino (approved):

PR #1087, opened 2025-02-18
Fix all Xsc tests to align Migration to Xray + Fix Watches failing tests

eranturgeman:
-----
This PR contains the following changes:
1) Fix in version validation call in all Xsc tests - now checking for the Xray version and decides if to address old Xsc service or the new one in Xray
2) Align all Xsc tests to operate in the same way using mocks
3) Fix TestXrayWatch that started failing sue to some change in the returned error message, now checking for containment and not exact error string (see picture)

Screenshot 2025-02-20 at 8 31 27


tests/xsc_test.go
```diff
@@ -25,29 +15,41 @@ func initXscTest(t *testing.T, minXscVersion string, minXrayVersion string) {
...
 	}
 
-	if minXscVersion != "" {
-		if err = clientUtils.ValidateMinimumVersion(clientUtils.Xsc, currentXscVersion, minXscVersion); err != nil {
+	afterMigration := true
+	if err = clientUtils.ValidateMinimumVersion(clientUtils.Xray, currentXrayVersion, utils.MinXrayVersionXscTransitionToXray); err != nil {
+		afterMigration = false
+	}
+
+	if afterMigration && minXrayVersion != "" {
```
attiasas, line 34:
```suggestion
	if minXrayVersion != "" {
```
why are we validating min xray version only if after migration?

tests/xsc_test.go
```diff
@@ -25,29 +15,41 @@ func initXscTest(t *testing.T, minXscVersion string, minXrayVersion string) {
...
+		if err = clientUtils.ValidateMinimumVersion(clientUtils.Xsc, currentXrayVersion, minXrayVersion); err != nil {
```
attiasas, line 35:
```suggestion
		if err = clientUtils.ValidateMinimumVersion(clientUtils.Xray, currentXrayVersion, minXrayVersion); err != nil {
```
Xray version is validated here, not xsc

attiasas (commented):
Nice, check out my comments

PR #1110, opened 2025-04-24
RTDEV-56127 - Support annotate command for release bundle

rimapol:
-----

### DIFF
--- a/lifecycle/manager.go
+++ b/lifecycle/manager.go
@@ -138,3 +138,8 @@ func (lcs *LifecycleServicesManager) IsReleaseBundleExist(rbName, rbVersion, pro
 	rbService := lifecycle.NewReleaseBundlesService(lcs.config.GetServiceDetails(), lcs.client)
 	return rbService.ReleaseBundleExists(rbName, rbVersion, projectKey)
 }
+
+func (lcs *LifecycleServicesManager) AnnotateReleaseBundle(params lifecycle.AnnotateOperationParams) error {
+	rbService := lifecycle.NewReleaseBundlesService(lcs.config.GetServiceDetails(), lcs.client)
+	return rbService.AnnotateReleaseBundle(params)
+}
--- a/lifecycle/services/annotate.go
+++ b/lifecycle/services/annotate.go
@@ -1,0 +1,207 @@
+package services
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/jfrog/jfrog-client-go/utils"
+	"github.com/jfrog/jfrog-client-go/utils/distribution"
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+	"github.com/jfrog/jfrog-client-go/utils/io/httputils"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+	"net/http"
+	"path"
+	"strings"
+)
+
+const (
+	Tag               = "tag"
+	PropertiesBaseApi = "api/storage"
+)
+
+type CommonPropParams struct {
+	Path      string `json:"path"`
+	Recursive bool   `json:"recursive"`
+}
+
+type ArtCommonParams struct {
+	Url string `json:"url"`
+}
+
+type AnnotateOperationParams struct {
+	RbTag          RbAnnotationTag
+	RbProps        RbAnnotationProps
+	RbDelProps     RbDelProps
+	RbDetails      ReleaseBundleDetails
+	QueryParams    CommonOptionalQueryParams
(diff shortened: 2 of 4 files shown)


rimapol:

lifecycle/services/annotate.go
```diff
@@ -0,0 +1,174 @@
...
+			return errorutils.CheckError(err2)
+		}
+		resp2, body2, err2 := rbs.client.SendPost(propsFullUrl, content2, &httpClientsDetails)
+		if err2 != nil {
+			return err2
+		}
+		if err = errorutils.CheckResponseStatusWithBody(resp2, body2, http.StatusCreated, http.StatusOK); err != nil {
+			return err
+		}
+	}
```
bhanurp, line 130: Couple of questions/comments here:
- there are 2 different for loops can this be changed to single for loop ?
- don't we have an API for taking multiple property bodies?
- can the variables be named with what it stores instead of content2, body2, resp2?
- reuse err instead of creating multiple err varaibles to ensure there is no shadowing.

bhanurp (changes requested):
- Update Readme about new API AnnotateReleaseBundle()
- Please remove unwanted new lines.

lifecycle/services/annotate.go
```diff
@@ -0,0 +1,174 @@
...
+
+	var initialStruct InitialProperties
+	err = json.Unmarshal(propsContent, &initialStruct)
+	if err != nil {
+		return errorutils.CheckError(err)
+	}
+
+	var propertyBodies []PropertyBody
+
+	propsFullUrl, err := utils.BuildUrl(artUrl, api, GetPropQueryParam(params.PropertyParams))
```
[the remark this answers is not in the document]
rimapol: @bhanurp yes, it's /artifactory

lifecycle/services/annotate.go
```diff
@@ -0,0 +1,207 @@
...
+
+func (rbs *ReleaseBundlesService) AnnotateReleaseBundle(params AnnotateOperationParams) error {
+	return rbs.annotateReleaseBundle(params)
+}
+
+func GetReleaseBundleSetTagApi(rbDetails ReleaseBundleDetails) string {
+	return path.Join(releaseBundleBaseApi, records, rbDetails.ReleaseBundleName, rbDetails.ReleaseBundleVersion, Tag)
+}
+
+func (rbs *ReleaseBundlesService) setTag(params AnnotateOperationParams, api string, httpClientsDetails httputils.HttpClientDetails) error {
```
bhanurp, line 64:
```suggestion
func (ao *AnnonateOperation) setTag(params AnnotateOperationParams, api string) error {
```
```
type AnnonateOperation struct {
 rtDetails *auth.ServiceDetails
 params AnnonateOperationParams
}
```
may be having a new struct for annotate operation will solve lot of problems with handling api, path, httpClientDetails makes more readable as well.
manager.go can have a new API on lifecycle service which take rtDetails as an input param.
WDYT?

lifecycle/manager.go
```diff
@@ -138,3 +138,8 @@ func (lcs *LifecycleServicesManager) IsReleaseBundleExist(rbName, rbVersion, pro
 	rbService := lifecycle.NewReleaseBundlesService(lcs.config.GetServiceDetails(), lcs.client)
 	return rbService.ReleaseBundleExists(rbName, rbVersion, projectKey)
 }
+
+func (lcs *LifecycleServicesManager) AnnotateReleaseBundle(params lifecycle.AnnotateOperationParams) error {
```
bhanurp, line 142:
```suggestion
func (lcs *LifecycleServicesManager) AnnotateReleaseBundle(params lifecycle.AnnotateOperationParams, artDetails auth.ServiceDetails, client *jfroghttpclient.JfrogHttpClient) error {
```
this method can define `NewAnnonateOPeration{}` with rtDetails, client etc so all required details are available in a single receiver instead of passing params

rimapol: @bhanurp I am not sure.

lifecycle/services/annotate.go
rimapol: @bhanurp I prefer not to change it. Because - 1) I see that in lifecycle/manager all methods are written by the same pattern. So, I'd like to keep it consistent. 2) I don't really know where from I get all needed data when I am in manager.go 3) I don't see any benefits from this change - there are no problems with handling api, path and httpClientDetails 4) We are luck of time, you didn't mention it in your previous code review, so, I suppose it's good enough. 
WDYT?

fluxxBot (approved):

PR #1126, opened 2025-06-05
Add catalog services

attiasas:
-----

Add **Catalog** services

* `Version` - returns `catalog` version
* `Enrich` - enrich a given CDX file with catalog information

### DIFF
--- a/artifactory/services/promote.go
+++ b/artifactory/services/promote.go
@@ -76,7 +76,7 @@ func (ps *PromoteService) BuildPromote(promotionParams PromotionParams) error {
 	if err != nil {
 		return err
 	}
-	
+
 	if err = errorutils.CheckResponseStatusWithBody(resp, body, http.StatusOK); err != nil {
 		return err
 	}
--- a/artifactory/services/utils/tests/catalog/server.go
+++ b/artifactory/services/utils/tests/catalog/server.go
@@ -21,7 +21,7 @@ func StartCatalogMockServerWithParams(t *testing.T, params MockServerParams) int
 	handlers := tests.HttpServerHandlers{}
 
 	handlers["/"] = http.NotFound
-	// Version handlers
+	// Version handlers (version is not available in Catalog, so we use ping endpoint)
 	handlers["/catalog/api/v1/system/ping"] = catalogGetVersionHandlerFunc(t, params)
 	// Enrich handlers
 	handlers["/catalog/api/v1/beta/cyclonedx/enrich"] = catalogEnrichHandler(t, params)
@@ -36,9 +36,6 @@ func StartCatalogMockServerWithParams(t *testing.T, params MockServerParams) int
 
 func catalogGetVersionHandlerFunc(t *testing.T, params MockServerParams) func(w http.ResponseWriter, r *http.Request) {
 	version := "1.0.0"
-	if !params.Alive {
-		version = ""
-	}
 	return func(w http.ResponseWriter, r *http.Request) {
 		if !params.Alive {
 			http.Error(w, "Catalog service is not available", http.StatusServiceUnavailable)
--- a/artifactory/services/utils/tests/catalog/server.go
+++ b/artifactory/services/utils/tests/catalog/server.go
@@ -1,0 +1,81 @@
+package catalog
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/CycloneDX/cyclonedx-go"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+	"github.com/jfrog/jfrog-client-go/utils/tests"
+	"github.com/stretchr/testify/assert"
+)
+
+type MockServerParams struct {
+	EnrichedVuln []cyclonedx.Vulnerability
+	Alive        bool
+}
+
+func StartCatalogMockServerWithParams(t *testing.T, params MockServerParams) int {
+	handlers := tests.HttpServerHandlers{}
+
+	handlers["/"] = http.NotFound
+	// Version handlers
+	handlers["/catalog/api/v1/system/ping"] = catalogGetVersionHandlerFunc(t, params)
+	// Enrich handlers
+	handlers["/catalog/api/v1/beta/cyclonedx/enrich"] = catalogEnrichHandler(t, params)
+
+	port, err := tests.StartHttpServer(handlers)
+	if err != nil {
+		log.Error(err)
+		os.Exit(1)
+	}
+	return port
+}
+
+func catalogGetVersionHandlerFunc(t *testing.T, params MockServerParams) func(w http.ResponseWriter, r *http.Request) {
+	version := "1.0.0"
+	if !params.Alive {
+		version = ""
+	}
+	return func(w http.ResponseWriter, r *http.Request) {
+		if !params.Alive {
+			http.Error(w, "Catalog service is not available", http.StatusServiceUnavailable)
+			return
+		}
+		if r.Method == http.MethodGet {
+			_, err := fmt.Fprint(w, version)
+			assert.NoError(t, err)
+			return
+		}
+		http.Error(w, "Invalid catalog request", http.StatusBadRequest)
+	}
+}
+
+func catalogEnrichHandler(t *testing.T, params MockServerParams) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if !params.Alive {
+			http.Error(w, "Catalog service is not available", http.StatusServiceUnavailable)
+			return
+		}
+		if r.Method == http.MethodPost {
+			// Read the BOM from the request body
+			bom := cyclonedx.NewBOM()
+			assert.NoError(t, cyclonedx.NewBOMDecoder(r.Body, cyclonedx.BOMFileFormatJSON).Decode(bom))
+			// Enrich the BOM with vulnerabilities
+			for _, vuln := range params.EnrichedVuln {
+				if bom.Vulnerabilities == nil {
+					bom.Vulnerabilities = &[]cyclonedx.Vulnerability{}
+				}
+				*bom.Vulnerabilities = append(*bom.Vulnerabilities, vuln)
+			}
+			// Encode the enriched BOM to JSON format and write it to the response
+			writer := cyclonedx.NewBOMEncoder(w, cyclonedx.BOMFileFormatJSON)
+			err := writer.Encode(bom)
+			assert.NoError(t, err)
+			return
+		}
+		http.Error(w, "Invalid enrich request", http.StatusBadRequest)
+	}
+}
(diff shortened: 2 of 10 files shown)


catalog/auth/catalogdetails.go
```diff
@@ -0,0 +1,45 @@
...
+		cs.Version, err = cs.getCatalogVersion()
+		if err != nil {
+			return "", err
+		}
+		log.Debug("JFrog Catalog version is:", cs.Version)
+	}
+	return cs.Version, nil
+}
+
+func (ds *catalogDetails) getCatalogVersion() (string, error) {
```
eranturgeman, line 31: If this is in use only once I can see no reason to split into 2 functions

eranturgeman (commented):
what about tests? I'd like to see tests for both version and enrich if possible

catalog/services/version.go
hadarshjfrog, line 1: please check with Catalog team - should we address catalog as a MT service differently on SaaS and On-Prem

catalog/auth/catalogdetails.go
hadarshjfrog, line 1: Could you please explain the difference between the usage of this file? why does it need its own auth and version?

catalog/manager.go
```diff
@@ -0,0 +1,61 @@
+package catalog
+
+import (
+	"github.com/CycloneDX/cyclonedx-go"
```
hadarshjfrog, line 4: how doesn't it add the `cyclonedx-go` to the `go.mod`?  when was it added before?

catalog/manager.go
```diff
@@ -0,0 +1,61 @@
...
+	client *jfroghttpclient.JfrogHttpClient
+	config config.Config
+	// Global reference to the provided project key, used for API endpoints that require it for authentication
+	scopeProjectKey string
+}
+
+// New creates a service manager to interact with Catalog
+func New(config config.Config) (*CatalogServicesManager, error) {
+	details := config.GetServiceDetails()
+	var err error
```
hadarshjfrog, line 22: Nuance - but I'd put this row either first, or just before the usage of it - a bit stuck there...

catalog/auth/catalogdetails.go
attiasas: to keep inline with other similar implementations.
One is a private and one is public, its probably to improve readability

catalog/manager.go
attiasas: it was indirect probably

catalog/manager.go
attiasas: its like that in other places, to keep inline with other services I think we should keep or change all

catalog/auth/catalogdetails.go
attiasas: the first one is a general implementation for `auth.ServiceDetails` interface, required for new service
the other is the actual service implementation

catalog/services/version.go
attiasas: I checked. If its new API, its OK to use `/catalog`

artifactory/services/utils/tests/catalog/server.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func StartCatalogMockServerWithParams(t *testing.T, params MockServerParams) int {
+	handlers := tests.HttpServerHandlers{}
+
+	handlers["/"] = http.NotFound
+	// Version handlers
+	handlers["/catalog/api/v1/system/ping"] = catalogGetVersionHandlerFunc(t, params)
+	// Enrich handlers
+	handlers["/catalog/api/v1/beta/cyclonedx/enrich"] = catalogEnrichHandler(t, params)
```
hadarshjfrog, line 27: beta? do they have an ETA to change it?
(obviously it's not up to us - but I suggest adding a comment that we use a beta version)

artifactory/services/utils/tests/catalog/server.go
```diff
@@ -0,0 +1,81 @@
...
+	EnrichedVuln []cyclonedx.Vulnerability
+	Alive        bool
+}
+
+func StartCatalogMockServerWithParams(t *testing.T, params MockServerParams) int {
+	handlers := tests.HttpServerHandlers{}
+
+	handlers["/"] = http.NotFound
+	// Version handlers
+	handlers["/catalog/api/v1/system/ping"] = catalogGetVersionHandlerFunc(t, params)
```
hadarshjfrog, line 25: no need for system/version also?

artifactory/services/utils/tests/catalog/server.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		log.Error(err)
+		os.Exit(1)
+	}
+	return port
+}
+
+func catalogGetVersionHandlerFunc(t *testing.T, params MockServerParams) func(w http.ResponseWriter, r *http.Request) {
+	version := "1.0.0"
+	if !params.Alive {
```
hadarshjfrog: why are we resetting the version in this case?

artifactory/services/utils/tests/catalog/server.go
hadarshjfrog: Ok - saw now they don't have - please refine the comment here too - that we use systm/ping instead of system/version

artifactory/services/promote.go
```diff
@@ -76,7 +76,7 @@ func (ps *PromoteService) BuildPromote(promotionParams PromotionParams) error {
 	if err != nil {
 		return err
 	}
-
```
hadarshjfrog, line 79: only whitespace difference? can we clear it?

artifactory/services/utils/tests/catalog/server.go
attiasas: not that I know of. I will ask

artifactory/services/utils/tests/catalog/server.go
attiasas: added

artifactory/services/promote.go
attiasas: it was the result of `go fmt ./...`

artifactory/services/utils/tests/catalog/server.go
attiasas: deleted

hadarshjfrog (approved):

artifactory/services/utils/tests/catalog/server.go
```diff
@@ -0,0 +1,78 @@
+package catalog
+
+import (
+	"fmt"
+	"net/http"
```
bhanurp, line 5: why not use "net/http/httptest" and move this to respective *_test.go file?

catalog/services/enricher.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/CycloneDX/cyclonedx-go"
+	"github.com/jfrog/jfrog-client-go/auth"
+	"github.com/jfrog/jfrog-client-go/http/jfroghttpclient"
+	"github.com/jfrog/jfrog-client-go/utils"
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+)
+
+const (
+	catalogEnrichApi              = "api/v1/beta/cyclonedx/enrich"
+	catalogMinVersionForEnrichApi = "1.0.0"
```
bhanurp, line 18: nit: where is this being used?

catalog/services/enricher.go
```diff
@@ -0,0 +1,72 @@
...
+func (es *EnrichService) enrich(bomJson []byte) ([]byte, error) {
+	httpDetails := es.CatalogDetails.CreateHttpClientDetails()
+	resp, body, err := es.client.SendPost(es.getUrlForEnrichApi(), bomJson, &httpDetails)
+	if err != nil {
+		return nil, errors.New("failed while attempting to enrich CycloneDX JSON BOM: " + err.Error())
+	}
+	if err = errorutils.CheckResponseStatusWithBody(resp, body, http.StatusOK); err != nil {
+		return nil, errors.New("got unexpected Catalog server response while attempting to enrich CycloneDX JSON BOM:\n" + err.Error())
+	}
+	if resp.StatusCode != http.StatusOK {
```
bhanurp: Is it possible the API can return 20X which is not 200 as success response?

catalog/services/version.go
```diff
@@ -0,0 +1,39 @@
...
+	httpDetails := vs.CatalogDetails.CreateHttpClientDetails()
+	resp, body, _, err := vs.client.SendGet(vs.CatalogDetails.GetUrl()+catalogPingApi, true, &httpDetails)
+	if err != nil {
+		return "", errors.New("failed while attempting to ping JFrog Catalog: " + err.Error())
+	}
+	if err = errorutils.CheckResponseStatusWithBody(resp, body, http.StatusOK); err != nil {
+		return "", errors.New("got unexpected server response while attempting to ping JFrog Catalog:\n" + err.Error())
+	}
+	// Since Catalog does not have a version endpoint, we return a hardcoded version.
+	return "1.0.0", nil
```
bhanurp: you probably wanted to use this constant `catalogMinVersionForEnrichApi` instead of "1.0.0"

utils/log/logger.go
```diff
@@ -52,20 +52,20 @@ const (
...
 }
 
 // Same as NewLoggerWithFlags, with log flags turned off.
-func NewLogger(logLevel LevelType, logToWriter io.Writer) *jfrogLogger {
+func NewLogger(logLevel LevelType, logToWriter io.Writer) *JfrogLogger {
 	return NewLoggerWithFlags(logLevel, logToWriter, 0)
 }
 
-type jfrogLogger struct {
+type JfrogLogger struct {
```
bhanurp, line 68: why is this change required?

bhanurp (changes requested):
Since catalog is wrapping a new API, may be it is a good idea to update README about how to use it

catalog/services/enricher.go
attiasas: In our security-cli repository while running the scan commands, used here:
https://github.com/jfrog/jfrog-cli-security/pull/478

catalog/services/enricher.go
attiasas: we don't accept any other status as 'success', so we only want to continue if we get 200

attiasas: removed duplicate check

utils/log/logger.go
attiasas: require to reference the inner `log.Logger` for each output.
We are starting to use a go plugin in our services, in order to align the logs we need access to create decorator with it, used here: https://github.com/jfrog/jfrog-cli-security/pull/479

catalog/services/version.go
attiasas: done

artifactory/services/utils/tests/catalog/server.go
attiasas: I was following the same impl used in: `artifactory/services/utils/tests/xray`

bhanurp: may not be part of this PR, but this can be improved by using net/http/httptest

PR #1141, opened 2025-06-27
Add get all permisison targets(fix #569)

### LINKED ISSUE #569: Support for get permission targets
**Is your feature request related to a problem? Please describe.**
The Artifactory REST API has a endpoint which returns a list of all permission targets. 
Unfortunately this library doesn't support this endpoint, you can only return a single permission target. 

**Describe the solution you'd like to see**
A method like `GetAllPermissionTargets`.

legnoh:
-----

I've implemented and added the function that resolves issue #569.
If you find anything missing or require further adjustments, please do not hesitate to let us know.

### DIFF
--- a/artifactory/services/permissiontarget.go
+++ b/artifactory/services/permissiontarget.go
@@ -62,6 +62,24 @@ func (pts *PermissionTargetService) Get(permissionTargetName string) (*Permissio
 	return permissionTarget, nil
 }
 
+func (pts *PermissionTargetService) GetAll() (*[]PermissionTargetParams, error) {
+	httpClientsDetails := pts.ArtDetails.CreateHttpClientDetails()
+	log.Info("Getting all permission target ...")
+	resp, body, _, err := pts.client.SendGet(pts.ArtDetails.GetUrl()+"api/v2/security/permissions", true, &httpClientsDetails)
+	if err != nil {
+		return nil, err
+	}
+	if err = errorutils.CheckResponseStatusWithBody(resp, body, http.StatusOK); err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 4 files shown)


artifactory/services/permissiontarget.go
```diff
@@ -62,6 +62,24 @@ func (pts *PermissionTargetService) Get(permissionTargetName string) (*Permissio
 	return permissionTarget, nil
 }
 
+func (pts *PermissionTargetService) GetAll() (*[]PermissionTargetParams, error) {
```
dortam888, line 65: This function looks like a code duplication of Get("") (Get with an empty string). 
I would suggest that you use the Get function with a condition on empty string to avoid duplication and allow maintainable changes.
Speaking of a change, from version 7.72.0 of Artifactory we moved to permissionV2 and we now have `GET /access/api/v2/permissions` (See [here](https://jfrog.com/help/r/jfrog-rest-apis/get-permissions)) 
It means that some permission targets will not be returned by this function, as they were created by access, and the security API will not find them. 

Please refactor the Get function to ask for the Artifactory version and use the new API if version is 7.72.0 or greater, and fallback to the old API otherwise

legnoh: I understand the idea of consolidating the logic into the `Get` function, and I will make that change.

However, I have some concerns about handling API switching based on versions. Given the current implementation trends, accessing `/access/api/v2/permissions` should be done using `accessManager`, and building it on top of `serviceManager` would be quite difficult.

Therefore, I propose treating this implementation as targeting the legacy API for now, and handling the `access` logic separately (e.g., in `access/services/permission.go`) at a later time. Does that sound reasonable?

PR #1165, opened 2025-08-20
Implement basic Get Violations API

attiasas:
-----

Add new Get Violations API to Xray manager following:
* https://jfrog.com/help/r/xray-rest-apis/get-violations

### DIFF
--- a/xray/services/utils/validations.go
+++ b/xray/services/utils/validations.go
@@ -1,0 +1,190 @@
+package utils
+
+import "math"
+
+const (
+	SecurityViolation        ViolationType = "Security"
+	LicenseViolation         ViolationType = "License"
+	OperationalRiskViolation ViolationType = "Operational_Risk"
+)
+
+type ViolationType string
+
+type ViolationsRequest struct {
+	Filters    *ViolationsFilters `json:"filters,omitempty"`
+	Pagination *PaginationOptions `json:"pagination,omitempty"`
+}
+
+type PaginationOptions struct {
+	// Valid values:  created, summary, severity, type, watcher_name, issue_id (Default: created)
+	OrderBy string `json:"order_by,omitempty"`
+	// Default if not provided: 25
+	Limit int `json:"limit,omitempty"`
+	// Default if not provided: 1
+	Offset int `json:"offset,omitempty"`
+	// Valid values: asc, desc (Default: For ordering by Severity: desc; for all other fields: asc)
+	Direction string `json:"direction,omitempty"`
+}
+
+type ViolationsFilters struct {
+	// Filtering the results for those included in the requested string in the description property
+	NameContains string `json:"name_contains,omitempty"`
+	// Add additional violation detail properties to the response (role name, policy name, description, remediation, and more)
+	IncludeDetails bool `json:"include_details,omitempty"`
+	// Filtering by the response for specific violation type. Valid values: Security, License, Operational_Risk
+	Type ViolationType `json:"violation_type,omitempty"`
+	// Filtering the results for those generated from the selected watch. Default: Any watch.
+	WatchName string `json:"watch_name,omitempty"`
+	// Filtering the results for those that their severity is equal or higher than min_severity.
+	// Valid values: Critical, High, Medium, Low, Information, Unknown (Note:  the values are listed in descending severity order)
+	MinSeverity Severity `json:"min_severity,omitempty"`
+	// Filter for violations created as of this time. Valid value:  A timestamp in RFC 3339 format
+	CreatedFrom string `json:"created_from,omitempty"`
+	// Filter for violations created up to this time. Valid value:  A timestamp in RFC 3339 format
+	CreatedUntil string `json:"created_until,omitempty"`
+	// Filter for violations resulting from the requested Issue ID.
+	// Valid values: strings representing the issue ID e.g: XRAY-94620, EXP-1552-00002, GPL-3.0, b1670bb2d3438da6213ed386577fd755bc, b8fdf85cab594e6a3717b4f182b07b
+	IssueId string `json:"issue_id,omitempty"`
+	// Filter for violations resulting from the requested CVE. Valid values:  a CVE standard identifier format <CVE-YYYY-NNNNNN>
+	CveId string `json:"cve_id,omitempty"`
+	// Filter for violations found in specific resources.
+	Resources ViolationResourceFilters `json:"resources,omitempty"`
+}
+
+type ViolationResourceFilters struct {
+	Artifacts        []ArtifactResourceFilter        `json:"artifacts,omitempty"`
+	Builds           []BuildResourceFilter           `json:"builds,omitempty"`
+	ReleaseBundles   []ReleaseBundleResourceFilter   `json:"release_bundles,omitempty"`
+	ReleaseBundlesV2 []ReleaseBundleV2ResourceFilter `json:"release_bundles_v2,omitempty"`
+	GitRepositories  []GitRepositoriesResourceFilter `json:"git_repositories,omitempty"`
+}
+
+type ArtifactResourceFilter struct {
+	Repository string `json:"repo"`
+	Path       string `json:"path"`
+}
+
+type BuildResourceFilter struct {
+	Name    string `json:"name"`
+	Number  string `json:"number"`
+	Project string `json:"project,omitempty"`
+}
+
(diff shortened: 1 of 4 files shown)


xray/services/utils/validations.go
```diff
@@ -0,0 +1,190 @@
...
+	Filters    *ViolationsFilters `json:"filters,omitempty"`
+	Pagination *PaginationOptions `json:"pagination,omitempty"`
+}
+
+type PaginationOptions struct {
+	// Valid values:  created, summary, severity, type, watcher_name, issue_id (Default: created)
+	OrderBy string `json:"order_by,omitempty"`
+	// Default if not provided: 25
+	Limit int `json:"limit,omitempty"`
+	// Default if not provided: 1
```
hadarshjfrog: why 1 and not 0?

xray/services/utils/validations.go
```diff
@@ -0,0 +1,190 @@
...
+
+type ViolationsRequest struct {
+	Filters    *ViolationsFilters `json:"filters,omitempty"`
+	Pagination *PaginationOptions `json:"pagination,omitempty"`
+}
+
+type PaginationOptions struct {
+	// Valid values:  created, summary, severity, type, watcher_name, issue_id (Default: created)
+	OrderBy string `json:"order_by,omitempty"`
+	// Default if not provided: 25
```
hadarshjfrog: isn't the default is 40?

xray/services/utils/validations.go
```diff
@@ -0,0 +1,190 @@
...
+	Limit int `json:"limit,omitempty"`
+	// Default if not provided: 1
+	Offset int `json:"offset,omitempty"`
+	// Valid values: asc, desc (Default: For ordering by Severity: desc; for all other fields: asc)
```
hadarshjfrog: where does it changes in case of Severity or everything else?

xray/services/utils/validations.go
```diff
@@ -0,0 +1,190 @@
...
+	Direction string `json:"direction,omitempty"`
+}
+
+type ViolationsFilters struct {
```
hadarshjfrog: I think that some of  filters are obvious (cveId, IssueId, MinSeverity From/Until) - but I'm thinking whether to keep it all of consistency sake...
WDYT?

xray/services/violations.go
```diff
@@ -0,0 +1,198 @@
...
+const (
+	violationsAPI = "api/v1/violations"
+)
+
+const (
+	Applicable    ApplicabilityStatus = "applicable"
+	NotApplicable ApplicabilityStatus = "not_applicable"
+	NotCovered    ApplicabilityStatus = "not_covered"
+	Undetermined  ApplicabilityStatus = "undetermined"
+
```
hadarshjfrog: why the break-line here? is there a logical separation between the statuses we can and can't get in SL? if so - mention it

xray/services/violations.go
```diff
@@ -0,0 +1,198 @@
...
+	Type     utils.ViolationType `json:"type"`
+	Watch    string              `json:"watch_name"`
+	Severity utils.Severity      `json:"severity"`
+	Created  string              `json:"created"`
+	// Additional violation details (depending on IncludeDetails flag, the violation type and configuration)
+	Id                   string              `json:"violation_id,omitempty"`
+	Description          string              `json:"description,omitempty"`
+	Summary              string              `json:"summary,omitempty"`
+	Provider             string              `json:"provider,omitempty"`
+	InfectedComponentIds []string            `json:"infected_components"`
```
hadarshjfrog: Is it intentional that `InfectedComponentIds` has `omitempty`, but `InfectedVerson` and `InfectedFilePaths` doesn't?

xray/services/violations.go
```diff
@@ -0,0 +1,198 @@
...
+
+type IgnoreRuleInfo struct {
+	Id             string `json:"id"`
+	Type           string `json:"ignore_rule_type"`
+	Author         string `json:"author"`
+	Created        string `json:"created"`
+	Notes          string `json:"notes"`
+	IsExpired      bool   `json:"is_expired"`
+	ExpirationDate string `json:"expires_at,omitempty"`
+	DeletedBy      string `json:"deleted_by,omitempty"`
```
hadarshjfrog, line 127: need to double check if we'll get the `Deleted...` fields (worth checking with Xray team)

xray/services/violations.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+type CveApplicabilityDetails struct {
+	ComponentId           string              `json:"component_id"`
+	VulnerableComponentId string              `json:"source_comp_id"`
+	CveId                 string              `json:"vulnerability_id"`
+	Status                ApplicabilityStatus `json:"result"`
+}
+
+type BaseJasDetails struct {
```
hadarshjfrog, line 178: I'm missing `fingerprints` - would we get it here? or is it parsed internally - worth checking with JAS team

xray/services/violations.go
```diff
@@ -0,0 +1,198 @@
...
+	Status       string         `json:"status"`
+	Severity     utils.Severity `json:"jfrog_severity,omitempty"`
+	Reason       string         `json:"description,omitempty"`
+	Id           string         `json:"id"`
```
hadarshjfrog: usually Id should be first...

xray/services/utils/validations.go
```diff
@@ -0,0 +1,190 @@
...
+	vr.Filters.IssueId = issueId
+	return vr
+}
+
+func (vr ViolationsRequest) FilterByCveId(cveId string) ViolationsRequest {
+	vr.Filters.CveId = cveId
+	return vr
+}
+
+func (vr ViolationsRequest) FilterByArtifacts(artifacts ...ArtifactResourceFilter) ViolationsRequest {
```
hadarshjfrog: why do all the array filters append instead of replacing?

xray/services/utils/validations.go
```diff
@@ -0,0 +1,190 @@
...
+type GitRepositoriesResourceFilter struct {
+	Name string `json:"name"`
+}
+
+func NewViolationsRequest() ViolationsRequest {
+	return ViolationsRequest{
+		Filters: &ViolationsFilters{},
+		Pagination: &PaginationOptions{
+			OrderBy:   "created",
+			Limit:     math.MaxInt,
```
hadarshjfrog: If we have pagination limit - replace with a limit of 40 (as a const of course) 

But I don't think we need limitations - so why adding the Pagination at all? is just intentionally to bring it all and be paginated by Xray? 
If so - I think the standard in Xray for "no pagination" is 1000.

xray/services/utils/validations.go
attiasas: based on the code from the Server...
Default usage with the struct is no limit (max int)

xray/services/utils/validations.go
attiasas: based on the code from the Server...
you can specify in the `OrderBy` attribute

xray/services/utils/validations.go
attiasas: This is a public API and in client-go repo, this is an option that is avaliable by the API...
The user can select what to use or filter with

xray/services/violations.go
attiasas: Following the structure definition in the server...
Also it's backward; there must be an infected component (id), but the component does not require a version or a file path

xray/services/utils/validations.go
attiasas: because you can filter by more than one thing...
no need to replace, just create new request...

xray/services/utils/validations.go
attiasas: this is not a UI. we want to "export" all violations...
If we are not passing the attribute, it will return the limit from the server causing multiple requests to get all info.
We rather call the API one time with "no limit"

xray/services/violations.go
attiasas: already reported that there are no fingerprints returned.... TBD (not sure if relevant for violation after there are generated already..., we can use the values from the matching vuln if we really wanted...)

xray/services/violations.go
attiasas: I will validate but I think this is information that only return from a similar but diff API (`/xray/api/v1/violations/ignored/{watch_name}`)

attiasas:
Continue in:
* https://github.com/jfrog/jfrog-client-go/pull/1186

PR #1177, opened 2025-09-08
Add Verbose log level

attiasas:
-----

Depends on:
* https://github.com/jfrog/build-info-go/pull/321

Add a new log level `Verbose` (lower level than DEBUG).
In addition, make some of the utils logs output in the new level to prevent big logs in our commands

### DIFF
--- a/artifactory/services/fspatterns/utils.go
+++ b/artifactory/services/fspatterns/utils.go
@@ -64,7 +64,7 @@ func filterFilesFunc(rootPath string, includeDirs, excludeWithRelativePath, pres
 			return false, err
 		}
 		if isExcludedByPattern {
-			log.Debug(fmt.Sprintf("The path '%s' is excluded", path))
+			log.Verbose(fmt.Sprintf("The path '%s' is excluded", path))
 			return false, nil
 		}
 
--- a/artifactory/services/fspatterns/utils.go
+++ b/artifactory/services/fspatterns/utils.go
@@ -75,7 +75,7 @@ func filterFilesFunc(rootPath string, includeDirs, excludeWithRelativePath, pres
 			}
 			// Check if the file size is within the limits
 			if !fileInfo.IsDir() && !sizeThreshold.IsSizeWithinThreshold(fileInfo.Size()) {
-				log.Debug(fmt.Sprintf("The path '%s' is excluded", path))
+				log.Verbose(fmt.Sprintf("The path '%s' is excluded", path))
 				return false, nil
 			}
 		}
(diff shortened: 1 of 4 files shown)


orto17 (approved):

artifactory/services/fspatterns/utils.go
```diff
@@ -75,7 +75,7 @@ func filterFilesFunc(rootPath string, includeDirs, excludeWithRelativePath, pres
 			}
 			// Check if the file size is within the limits
 			if !fileInfo.IsDir() && !sizeThreshold.IsSizeWithinThreshold(fileInfo.Size()) {
```
orto17, line 77: this is not related to the current change, but if a file is excluded because of its size, shouldn't be a different log to indicate the exclusion reason?

attiasas: added more details to the exclude log

utils/log/logger.go
```diff
@@ -122,10 +125,11 @@ var prefixStyles = map[LevelType]struct {
 	color    color.Color
 	emoji    string
 }{
-	DEBUG: {logLevel: "Debug", color: color.Cyan},
-	INFO:  {logLevel: "Info", emoji: "🔵", color: color.Blue},
-	WARN:  {logLevel: "Warn", emoji: "🟠", color: color.Yellow},
-	ERROR: {logLevel: "Error", emoji: "🚨", color: color.Red},
+	VERBOSE: {logLevel: "Verbose", color: color.Gray},
```
cyrilc-pro, line 128: Does the length of `VERBOSE` string mess with way the logs look? Is there some padding added to log level to align fields?

cyrilc-pro (approved):

utils/log/logger.go
attiasas: Not implemented..., but 'Error'/'Debug' also has 5 char and not 4 (in our case, it's more noticeable since it's 7 chars)
IMO, not relevant to this PR and should not be any different to the users since most of them will never see it.

go.mod
```diff
@@ -56,6 +56,6 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-// replace github.com/jfrog/build-info-go => github.com/jfrog/build-info-go v1.8.9-0.20250203111011-4ff16d3d42be
+replace github.com/jfrog/build-info-go => github.com/jfrog/build-info-go v1.8.9-0.20250910075402-a0c1a53e465a
```
ehl-jf, line 59: @attiasas  Since we now target master directly, we dont allow replace anymore. If you need a new version of build-info-go, please make sure its released and use that version instead of the replace.

attiasas:
continue in:
* https://github.com/jfrog/jfrog-client-go/pull/1188

PR #1186, opened 2025-09-14
Xray Get Violations API

attiasas:
-----

Add new Get Violations API to Xray manager following:
* https://jfrog.com/help/r/xray-rest-apis/get-violations

### DIFF
--- a/apptrust/services/application.go
+++ b/apptrust/services/application.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	applicationDetailsAPI = "apptrust/api/v1/applications"
+	applicationDetailsAPI = "api/v1/applications"
 )
 
 type ApplicationService struct {
--- a/apptrust/services/application_test.go
+++ b/apptrust/services/application_test.go
@@ -89,7 +89,7 @@ func TestApplicationService_GetApplicationDetails_Forbidden(t *testing.T) {
 func TestApplicationService_GetApplicationDetails_InvalidJSON(t *testing.T) {
 	requestNum := 0
 	handlerFunc := func(w http.ResponseWriter, r *http.Request) {
-		expectedURI := "/apptrust/api/v1/applications/" + mockApplicationKey
+		expectedURI := "/api/v1/applications/" + mockApplicationKey
 		if r.RequestURI == expectedURI {
 			assert.Equal(t, "GET", r.Method)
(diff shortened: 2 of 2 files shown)


xray/services/utils/violations.go
```diff
@@ -0,0 +1,190 @@
...
+type GitRepositoriesResourceFilter struct {
+	Name string `json:"name"`
+}
+
+func NewViolationsRequest() ViolationsRequest {
+	return ViolationsRequest{
+		Filters: &ViolationsFilters{},
+		Pagination: &PaginationOptions{
+			OrderBy:   "created",
+			Limit:     math.MaxInt,
```
hadarshjfrog, line 93: Should we sent maxInt? 
What happens if we omit it? and if won't it'll be 40? 1000?

xray/services/violations.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	response = &ViolationsResponse{}
+	if err = json.Unmarshal(body, response); err != nil {
+		return nil, errorutils.CheckErrorf("couldn't parse JFrog Xray server violations response: %s", err.Error())
+	}
+	return response, nil
+}
+
+type ViolationsResponse struct {
```
hadarshjfrog, line 68: I'm not sure about the `omitempty` here

xray/services/violations.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+type ViolationPolicy struct {
+	PolicyName string `json:"policy"`
+	Rule       string `json:"rule"`
+
+	FailBuild         bool `json:"is_build_failed,omitempty"`
+	FailPullRequest   bool `json:"fail_pull_request,omitempty"`
+	SkipNotApplicable bool `json:"is_skip_not_applicable,omitempty"`
+
```
hadarshjfrog: why the split here? 
and maybe split between between the `omitempty` and without?

xray/services/utils/violations.go
attiasas: there is default value, after consulting the way to get all is to pass max in limit

xray/services/violations.go
attiasas: copy paste from the API at the server

xray/services/violations.go
attiasas: removed space

PR #1197, opened 2025-09-15
Add support for apptrust manager and getApplicationDetail API

dortam888:
-----

Add Support for AppTrust Client Manager in client.
Add GetApplicationDetails API to go-client.
Add README section of AppTrust manager

### DIFF
--- a/apptrust/manager.go
+++ b/apptrust/manager.go
@@ -6,19 +6,12 @@ import (
 	"github.com/jfrog/jfrog-client-go/http/jfroghttpclient"
 )
 
-// ApptrustServiceManagerInterface defines the interface for ApptrustServicesManager to enable mocking in tests
-type ApptrustServiceManagerInterface interface {
-	Client() *jfroghttpclient.JfrogHttpClient
-	GetApplicationDetails(applicationKey string) (*services.Application, error)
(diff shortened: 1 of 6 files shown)


apptrust/manager.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/jfrog/jfrog-client-go/config"
+	"github.com/jfrog/jfrog-client-go/http/jfroghttpclient"
+)
+
+type ApptrustServicesManager struct {
+	client *jfroghttpclient.JfrogHttpClient
+	config config.Config
+}
+
+func New(config config.Config) (*ApptrustServicesManager, error) {
```
mnsboev, line 14: Please add interface for ApptrustServiceManager to make it mockable and return here this interface

mnsboev (changes requested):

apptrust/manager.go
dortam888: While I agree that this can be mockable for unit tests, 
I **disagree** with this change comment as it is written in this specific format in all the other services and will be used by other services, not only ours.
If we decide to change the format, it should be a refactor to all the services; otherwise, there will be confusion when using this specific service in other dependent services.
This change should not be made by us but by the maintainers.

RemiBou: also interface should be created by consumer, not provider, if mock is needed

PR #1199, opened 2025-09-17
support for direct-download cmd 

reshmifrog:
-----
Problem: We were experiencing an issue where artifacts are not being resolved according to the defined resolution order in a virtual repository when using the JFrog CLI.

Solution: Added support for direct download command to download artifacts bypassing the aql

### DIFF
--- a/artifactory/services/directdownload.go
+++ b/artifactory/services/directdownload.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -355,7 +356,7 @@ func (dds *DirectDownloadService) getFilesFromDirectory(repo, dirPath string, pa
 	}
 
 	for _, file := range files {
-		filePath := filepath.Join(dirPath, file.Name)
+		filePath := path.Join(dirPath, file.Name)
 		filesToDownload = append(filesToDownload, filePath)
 	}
 
@@ -377,7 +378,7 @@ func (dds *DirectDownloadService) collectAllFilesRecursively(repo, basePath stri
 		}
 
 		for _, item := range items {
-			itemPath := filepath.Join(currentDir, item.Name)
+			itemPath := path.Join(currentDir, item.Name)
 			if item.Folder {
 				// Add subdirectory to process
 				dirsToProcess = append(dirsToProcess, itemPath)
@@ -412,7 +413,7 @@ func (dds *DirectDownloadService) getFilesMatchingPattern(repo, pattern string, 
 
 	for _, file := range files {
 		if matched, _ := filepath.Match(filePattern, file.Name); matched {
-			filePath := filepath.Join(dir, file.Name)
+			filePath := path.Join(dir, file.Name)
 			if !dds.isExcluded(filePath, params.GetExclusions()) {
 				filesToDownload = append(filesToDownload, filePath)
 			}
@@ -439,11 +440,11 @@ func (dds *DirectDownloadService) collectFilesRecursively(repo, basePath, filePa
 		for _, item := range items {
 			if item.Folder {
 				// Add subdirectory to process
-				dirsToProcess = append(dirsToProcess, filepath.Join(currentDir, item.Name))
+				dirsToProcess = append(dirsToProcess, path.Join(currentDir, item.Name))
 			} else {
 				// Check if file matches pattern
 				if matched, _ := filepath.Match(filePattern, item.Name); matched {
-					filePath := filepath.Join(currentDir, item.Name)
+					filePath := path.Join(currentDir, item.Name)
 					if !dds.isExcluded(filePath, params.GetExclusions()) {
 						*result = append(*result, filePath)
 					}


artifactory/manager.go
```diff
@@ -290,6 +298,17 @@ func (sm *ArtifactoryServicesManagerImp) DownloadFilesWithSummary(params ...serv
 	return downloadService.DownloadFiles(params...)
 }
 
+func (sm *ArtifactoryServicesManagerImp) DirectDownloadFiles(params ...services.DirectDownloadParams) (totalDownloaded, totalFailed int, err error) {
```
fluxxBot, line 307: Can we rename the function with something meaningful? like downloadFilesWithApi or downloadFilesWithoutAql or whatever seems for appropriate and self explainatory.

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,363 @@
...
+
+		if dds.containsWildcards(artifactPath) {
+			count, failed, err := dds.handleWildcardDownload(repo, artifactPath, &params)
+			summary.TotalSucceeded += count
+			summary.TotalFailed += failed
+			if err != nil {
+				log.Error(err)
+			}
+		} else {
+			success, err := dds.downloadSingleFile(repo, artifactPath, &params)
```
fluxxBot: why is there specific logic to download a single file? can't it be downloaded with patterns? add comments to make this complex conditions readable

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,363 @@
...
+			continue
+		}
+
+		fileName := strings.TrimPrefix(child.Uri, "/")
+		matched, err := filepath.Match(filePattern, fileName)
+		if err != nil {
+			return downloadCount, failCount, err
+		}
+
+		if matched {
```
fluxxBot: what are we matching here? can this be done before downloading, so as to download only files which match the given criteria?

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,363 @@
...
+			if !dds.isExcluded(filepath.Join(dir, fileName), params.GetExclusions()) {
+				filePath := filepath.Join(dir, fileName)
+				success, err := dds.downloadSingleFile(repo, filePath, params)
```
fluxxBot: downloading each file again and again recursively will create performance implications, check if this can be done via original api directly?

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,363 @@
...
+			sourcePath = strings.TrimPrefix(sourcePath, rtUrl)
+			if !strings.HasPrefix(sourcePath, "/") {
+				sourcePath = "/" + sourcePath
+			}
+		}
+
+		fileTransferDetails := clientutils.FileTransferDetails{
+			SourcePath: sourcePath,
+			TargetPath: localPath,
+			RtUrl:      rtUrl,
```
fluxxBot: by any chance can this also include sha256 as well? original struct has a placeholder for this

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,363 @@
...
+					failCount++
+				}
+			}
+		}
+	}
+
+	return downloadCount, failCount, nil
+}
+
+func (dds *DirectDownloadService) validateChecksum(downloadUrl, localPath string) error {
```
fluxxBot: how are we verifying checksum for pattern based file downloading ?

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,363 @@
...
+	if err := json.Unmarshal(body, &fileInfo); err != nil {
+		return errorutils.CheckError(err)
+	}
+
+	localFileDetails, err := fileutils.GetFileDetails(localPath, true)
+	if err != nil {
+		return err
+	}
+
+	if fileInfo.Checksums.Md5 != "" && localFileDetails.Checksum.Md5 != fileInfo.Checksums.Md5 {
```
fluxxBot: why are we explicitly checking for empty md5, sha1 and sha256? anyway if it's empty the comparison would fail

artifactory/services/directdownloadparams.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (ddp *DirectDownloadParams) IsQuiet() bool {
+	return ddp.Quiet
+}
+
+func (ddp *DirectDownloadParams) SetQuiet(quiet bool) {
+	ddp.Quiet = quiet
+}
+
+func NewDirectDownloadParams() *DirectDownloadParams {
```
fluxxBot: check if this default values are same as the download api

artifactory/services/directdownload.go
reshmifrog: Wildcards require pattern matching that direct API doesn't support , single files can use direct api, wildcards need storage api listing and pattern matching.

artifactory/services/directdownload.go
reshmifrog: Direct API doesn't support pattern matching, so individual downloads are required

artifactory/services/directdownload.go
reshmifrog: It's checking if filename matches the wildcard pattern after listing directory contents
and cannot be done before downloading because we need storage api response first

artifactory/services/directdownload.go
reshmifrog: artifactory might not always provide all checksums , some repositories might only have MD5/SHA1 but not SHA256

artifactory/services/directdownload.go
reshmifrog: Each file that matches the pattern is downloaded and verified, which includes full checksum validation (MD5, SHA1, SHA256) .

artifactory/manager.go
```diff
@@ -275,6 +275,14 @@ func (sm *ArtifactoryServicesManagerImp) initDownloadService() *services.Downloa
 	return downloadService
 }
 
+func (sm *ArtifactoryServicesManagerImp) initDirectDownloadService() *services.DirectDownloadService {
```
fluxxBot, line 284: nit: I think even this name needs to be changed

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+
+	if err := json.Unmarshal(body, &storageInfo); err != nil {
+		return 0, 0, err
+	}
+
+	downloadCount := 0
+	failCount := 0
+
+	for _, child := range storageInfo.Children {
```
fluxxBot: I see this is a for loop downloading files being matched by the wildcard pattern, by any chance is it possible to download all the children files parallely, i.e. using goroutines?

artifactory/services/directdownload.go
fluxxBot: both the conditions makes the same decision

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,387 @@
...
+	summary, err := dds.performDirectDownload(downloadParams...)
+	if err != nil {
+		return 0, 0, err
+	}
+	return summary.TotalSucceeded, summary.TotalFailed, nil
+}
+
+// DownloadFilesWithoutAQLWithSummary downloads files directly from Artifactory without using AQL queries
+// and gives detailed information about each file transfer.
+func (dds *DirectDownloadService) DownloadFilesWithoutAQLWithSummary(downloadParams ...DirectDownloadParams) (operationSummary *utils.OperationSummary, err error) {
```
fluxxBot: I see that you've made 3 api calls :-
1. to list all the files in the directory
2. to download the actual file
3. to fetch the fileInfo from server to validate checksum

is there by any chance we can reduce this?
effectively can we verify if downloading the actual file also gets us the file metadata from the server so as to reduce a network overhead?

fluxxBot (changes requested):

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,387 @@
...
+	if dds.saveSummary && dds.filesTransfersWriter != nil {
+		if err = dds.filesTransfersWriter.Close(); err != nil {
+			return nil, err
+		}
+		filePath := dds.filesTransfersWriter.GetFilePath()
+		log.Debug("Creating content reader from file:", filePath)
+
+		summary.TransferDetailsReader = content.NewContentReader(filePath, content.DefaultKey)
+	} else {
+		log.Debug("Not creating content reader - saveSummary:", dds.saveSummary, "filesTransfersWriter:", dds.filesTransfersWriter != nil)
```
bhanurp: is this required? else block can be removed.

artifactory/services/directdownloadparams.go
```diff
@@ -0,0 +1,104 @@
...
+	return &DirectDownloadParams{
+		CommonParams:       &utils.CommonParams{},
+		MinSplitSizeMB:     5120,
+		SplitCount:         3,
+		Retries:            3,
```
bhanurp: Retries is set in directdownload.go in cli-artifactory then why is it required to add a default value here?

artifactory/services/directdownloadparams.go
```diff
@@ -0,0 +1,104 @@
...
+		RetryWaitMilliSecs: 0,
```
bhanurp: this is redundant default value should be 0.

artifactory/services/directdownloadparams.go
```diff
@@ -0,0 +1,104 @@
...
+	return ddp.Quiet
+}
+
+func (ddp *DirectDownloadParams) SetQuiet(quiet bool) {
+	ddp.Quiet = quiet
+}
+
+func NewDirectDownloadParams() *DirectDownloadParams {
+	return &DirectDownloadParams{
+		CommonParams:       &utils.CommonParams{},
```
bhanurp: if at least one of the common param supported for direct download then we need to include this otherwise it is not required.

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,387 @@
...
+			// Handle patterns like "*.zip" or "test?.jar" by listing the directory
+			// and downloading matching files one by one
+			count, failed, err := dds.handleWildcardDownload(repo, artifactPath, &params)
+			summary.TotalSucceeded += count
+			summary.TotalFailed += failed
+			if err != nil {
+				log.Error(err)
+			}
+		} else {
+			// directly download the file in case wildcards are not present
```
bhanurp: nit pick: comments should start with upper case
```suggestion
			// Directly download the file in case wildcards are not present
```

artifactory/services/directdownload.go
bhanurp: may be else blog can be removed

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,387 @@
...
+func (dds *DirectDownloadService) SetArtifactoryDetails(artDetails auth.ServiceDetails) {
+	dds.artDetails = &artDetails
+}
+
+func (dds *DirectDownloadService) SetSaveSummary(saveSummary bool) {
+	dds.saveSummary = saveSummary
+}
+
+// DownloadFilesWithoutAQL downloads files directly from Artifactory without using AQL queries.
+func (dds *DirectDownloadService) DownloadFilesWithoutAQL(downloadParams ...DirectDownloadParams) (totalDownloaded, totalFailed int, err error) {
```
bhanurp: no need of named params since data is extracted and returned.
```suggestion
func (dds *DirectDownloadService) DownloadFilesWithoutAQL(downloadParams ...DirectDownloadParams) (int, int, error) {
```

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,387 @@
...
+			success, err := dds.downloadSingleFile(repo, artifactPath, &params)
+			switch {
+			case err != nil:
+				log.Error(err)
+				summary.TotalFailed++
+			case success:
+				summary.TotalSucceeded++
+			default:
+				summary.TotalFailed++
+			}
```
bhanurp: this is not readable better to remove switch case 

```suggestion
			// Directly download the file if wildcards are not present
success, err := dds.downloadSingleFile(repo, artifactPath, &params)
if err != nil {
	log.Error(err)
	summary.TotalFailed++
}

if success {
	summary.TotalSucceeded++
} else {
	summary.TotalFailed++
}
```

artifactory/services/directdownload.go
```diff
@@ -0,0 +1,387 @@
...
+
+		summary.TransferDetailsReader = content.NewContentReader(filePath, content.DefaultKey)
+	} else {
+		log.Debug("Not creating content reader - saveSummary:", dds.saveSummary, "filesTransfersWriter:", dds.filesTransfersWriter != nil)
+	}
+
+	return summary, nil
+}
+
+func (dds *DirectDownloadService) parsePattern(pattern string) (repo, artifactPath string, err error) {
```
bhanurp:
```suggestion
func (dds *DirectDownloadService) parsePattern(pattern string) (string, string, error) {
```

artifactory/manager.go
reshmifrog: this is as per the design document

PR #1257, opened 2025-11-03
fix: new fields and fixes for Terraform, Cargo, Nuget, Helm, Ansible

richardlt:
This PR add some missing fields on local and remote repository structs. The Ansible implmentation was also added.
Another changes that is introduced here is removing some omitempty tags. Without removing this tags it seems impossible to empty some fields when updating the repositories config.

-----

### DIFF
--- a/tests/artifactorylocalrepository_test.go
+++ b/tests/artifactorylocalrepository_test.go
@@ -649,6 +649,7 @@ func localTerraformTest(t *testing.T) {
 	tlp := services.NewTerraformLocalRepositoryParams()
 	tlp.Key = repoKey
 	setLocalRepositoryBaseParams(&tlp.LocalRepositoryBaseParams, false)
+	setTerraformRepositoryParams(&tlp.TerraformRepositoryParams, false)
 
 	err := testsCreateLocalRepositoryService.Terraform(tlp)
 	if !assert.NoError(t, err, "Failed to create "+repoKey) {
@@ -658,6 +659,7 @@ func localTerraformTest(t *testing.T) {
 	validateRepoConfig(t, repoKey, tlp)
 
 	setLocalRepositoryBaseParams(&tlp.LocalRepositoryBaseParams, true)
+	setTerraformRepositoryParams(&tlp.TerraformRepositoryParams, true)
 
 	err = testsUpdateLocalRepositoryService.Terraform(tlp)
 	if assert.NoError(t, err, "Failed to update "+repoKey) {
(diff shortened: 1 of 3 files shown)


richardlt:
Work in progress to adapt unit tests.

RemiBou (approved):

RemiBou:
please check CI failures

artifactory/services/remoterepository.go
```diff
@@ -371,7 +383,7 @@ func NewGradleRemoteRepositoryParams() GradleRemoteRepositoryParams {
 
 type HelmRemoteRepositoryParams struct {
 	RemoteRepositoryBaseParams
-	ChartsBaseUrl                string   `json:"chartsBaseUrl,omitempty"`
+	ChartsBaseUrl                string   `json:"chartsBaseUrl"` // do not set omitempty to be able to empty the value when updating
```
fluxxBot, line 386: if let's say this used for updating, are we making sure any empty value sent to the server is not updated in the database?

tests/artifactoryvirtualrepository_test.go
```diff
@@ -675,8 +675,6 @@ func virtualTerraformTest(t *testing.T) {
 	setCacheVirtualRepositoryParams(&tvp.CommonCacheVirtualRepositoryParams, true)
 
 	err = testsUpdateVirtualRepositoryService.Terraform(tvp)
-	assert.NoError(t, err, "Failed to update "+repoKey)
```
fluxxBot, line 678: isn't this assertion valid? Is it possible to provide proper comment in the codebase why it was decided to be removed?

tests/artifactoryvirtualrepository_test.go
```diff
@@ -675,8 +675,6 @@ func virtualTerraformTest(t *testing.T) {
...
-	validateRepoConfig(t, repoKey, tvp)
```
fluxxBot, line 679: same

tests/artifactoryvirtualrepository_test.go
richardlt: Removed this two lines because there are the same assertions at line 680.

artifactory/services/remoterepository.go
richardlt: I'm not sure about the implementation in the Artifactory API since I don't have the source for it. But in the client updating local and remote repositories config will result in a call to the method https://github.com/jfrog/jfrog-client-go/blob/master/artifactory/services/repository.go#L28 that will marshal the struct (remove the empty values from the body) and send a POST request. Maybe the POST is implemented like a PATCH on the server and ignore missing fields.

richardlt: Also what strange in that `performRequest` function is that the update will run a POST while the create will do a PUT.

PR #1265, opened 2025-11-18
Fix an issue that evidence encoded url twice resulting in a wrong url

dortam888:
-----

Evidence encoded url twice resulting in not finding subjects when contains decoded chars as we encoded again the % (For example space became %20 and %2520)

### DIFF
--- a/evidence/evidence_test.go
+++ b/evidence/evidence_test.go
@@ -199,25 +199,25 @@ func TestUploadEvidence_URLEncodingFix(t *testing.T) {
 		{
 			name:              "Build name with spaces (original bug case)",
 			subjectUri:        "ssfpoc-build-info/SSF Demo Electron/2-1762810453125.json",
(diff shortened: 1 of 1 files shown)


alenon (approved):

evidence/evidence_test.go
```diff
@@ -186,3 +186,102 @@ func createErrorHandlerFuncVersion(t *testing.T) (http.HandlerFunc, *int) {
...
+			var capturedRequestPath string
+			var capturedRawURL string
+
+			// Create a mock handler that captures the actual request path
+			mockHandler := func(w http.ResponseWriter, r *http.Request) {
+				switch {
+				case r.URL.Path == "/api/v1/system/version":
+					w.WriteHeader(http.StatusOK)
+
+				case strings.HasPrefix(r.URL.Path, "/api/v1/subject/"):
```
RemiBou: no "evidence" prefix in the url ?

dortam888: Evidence prefix is assembled in evidence-cli when creating the client. 
Its a good point though, I will mock it in these tests as well as it's more specific and gives better meaning to the tests.

PR #1272, opened 2025-12-04
Git Integration POST Request

orto17:
-----

### DIFF
--- a/xsc/manager.go
+++ b/xsc/manager.go
@@ -90,7 +90,6 @@ func (sm *XscServicesManager) GetConfigProfileByUrl(_ string) (*services.ConfigP
 
 // SendGitIntegrationEvent sends a git integration event to xray
 func (sm *XscServicesManager) SendGitIntegrationEvent(event services.GitIntegrationEvent, xrayVersion string) error {
-	eventService := services.NewAnalyticsEventService(sm.client)
-	eventService.XscDetails = sm.config.GetServiceDetails()
-	return eventService.SendGitIntegrationEvent(event, xrayVersion)
+	// Empty implementation required for alignment with interface, implemented only at the new service inside the Xray service
+	return nil
 }
--- a/xsc/services/analytics.go
+++ b/xsc/services/analytics.go
@@ -114,7 +114,7 @@ func (vs *AnalyticsEventService) AddGeneralEvent(event XscAnalyticsGeneralEvent,
 // SendGitIntegrationEvent sends a POST request to the /git_integration_event endpoint
 func (vs *AnalyticsEventService) SendGitIntegrationEvent(event GitIntegrationEvent, xrayVersion string) error {
 	if err := utils.ValidateMinimumVersion(utils.Xray, xrayVersion, xscutils.MinXrayVersionGitIntegrationEvent); err != nil {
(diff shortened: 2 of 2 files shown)


xsc/services/analytics.go
```diff
@@ -16,6 +16,7 @@ import (
 const (
 	AnalyticsMetricsMinXscVersion = "1.7.1"
 	xscEventApi                   = "event"
+	gitIntegrationEventApi        = "git_integration_event"
```
attiasas, line 19: API with `_`? is it ok?

xsc/services/analytics.go
```diff
@@ -100,6 +111,25 @@ func (vs *AnalyticsEventService) AddGeneralEvent(event XscAnalyticsGeneralEvent,
 	return response.MultiScanId, errorutils.CheckError(err)
 }
 
+// SendGitIntegrationEvent sends a POST request to the /git_integration_event endpoint
+func (vs *AnalyticsEventService) SendGitIntegrationEvent(event GitIntegrationEvent, xrayVersion string) error {
+	if err := utils.ValidateMinimumVersion(utils.Xray, xrayVersion, xscutils.MinXrayVersionGitIntegrationEvent); err != nil {
+		return fmt.Errorf("git integration event requires minimum Xray version %s: %w", xscutils.MinXrayVersionGitIntegrationEvent, err)
```
attiasas: error generated from `ValidateMinimumVersion` already contains some of the info you are displaying.
```
You are using %s version %s, while this operation requires version %s or higher.
```
consider changing the err

xsc/services/analytics.go
```diff
@@ -201,6 +231,16 @@ type PullRequestContext struct {
...
 }
 
+type GitIntegrationEvent struct {
+	EventType     string `json:"event_type"`
+	GitProvider   string `json:"git_provider"`
+	GitOwner      string `json:"git_owner"`
+	GitRepository string `json:"git_repository"`
+	GitBranch     string `json:"git_branch"`
+	EventStatus   string `json:"event_status"`
+	FailureReason string `json:"failure_reason,omitempty"`
```
attiasas, line 241: if we have this values, we don't need an update func for an event of this type?

xsc/manager.go
```diff
@@ -87,3 +87,10 @@ func (sm *XscServicesManager) GetConfigProfileByUrl(_ string) (*services.ConfigP
 	// Empty implementation required for alignment with interface, implemented only at the new service inside the Xray service
 	return nil, nil
 }
+
+// SendGitIntegrationEvent sends a git integration event to xray
+func (sm *XscServicesManager) SendGitIntegrationEvent(event services.GitIntegrationEvent, xrayVersion string) error {
+	eventService := services.NewAnalyticsEventService(sm.client)
+	eventService.XscDetails = sm.config.GetServiceDetails()
+	return eventService.SendGitIntegrationEvent(event, xrayVersion)
```
attiasas: You can leave it empty, similar to `GetConfigProfileByUrl`
This API is not supported in the depracated (multi tanent APIs) only at the new service in Xray

xsc/services/analytics.go
orto17: yes

xsc/services/analytics.go
orto17: no, we only send on event at the end of the operation, regardless of failure/success

PR #1276, opened 2025-12-11
Fix: Temp Directory Cleanup Continues on Permission Errors

agrasth:
-----
# Fix: Temp Directory Cleanup Continues on Permission Errors

## Problem

When the JFrog CLI runs cleanup of old temporary files in `/tmp`, if it encounters a file owned by a different user (permission error), it stops immediately and does not continue to delete the remaining files that the current user has permission to delete.

### Example Scenario

Given 3 temp files in `/tmp`:

| Filename | Owner |
|----------|-------|
| `jfrog.cli.temp.-000000-12345` | user-2 |
| `jfrog.cli.temp.-111111-12345` | user-1 |
| `jfrog.cli.temp.-999999-12345` | user-2 |

When `user-2` runs `jf -v`:
- **Before fix**: File `-000000-` is deleted, permission error on `-111111-`, cleanup **stops** - file `-999999-` is **not deleted**
- **After fix**: File `-000000-` is deleted, permission error on `-111111-` is **collected**, file `-999999-` is deleted, then error is reported

## Root Cause

In `utils/io/fileutils/temp.go`, the `CleanOldDirs()` function returned immediately upon encountering any error during file deletion:

```go
// Before fix - returns immediately on first error
if err = RemovePath(filePath); err != nil {
    return fmt.Errorf("Cannot remove path: %s due to: %s", filePath, err.Error())
}
```

## Solution

Modified `CleanOldDirs()` to collect all deletion failures and continue processing remaining files. At the end, it reports all failures in an aggregated error message:

```go
// After fix - collects errors and continues
var failedDeletions []string

for _, file := range files {
    // ... 
    if err = RemovePath(filePath); err != nil {
        // Collect deletion errors and continue processing
        failedDeletions = append(failedDeletions,
            fmt.Sprintf("%s (%v)", filePath, err))
        continue
    }
}

// At the end, return aggregated error message for all failures
if len(failedDeletions) > 0 {
    return fmt.Errorf("failed to cleanup %d file(s): %s",
        len(failedDeletions), strings.Join(failedDeletions, "; "))
}
```

## Changes

- **File**: `utils/io/fileutils/temp.go`
  - Modified `CleanOldDirs()` to use `continue` instead of `return` on errors
  - Added `failedDeletions` slice to collect all failures
  - Returns aggregated error message at the end with count of failed files

### Unit Test Added

Added `TestCleanOldDirsContinuesOnError` in `temp_test.go` to verify:
1. Files that can be deleted are deleted
2. Cleanup continues after encountering errors
3. All failures are reported in the final error message

### Docker-based Integration Test

Reproduced the issue and verified the fix using Docker:

**Before Fix (jf 2.85.0):**
```
=== Files before jf -v ===
-rw-r--r--    1 guest    users    /tmp/jfrog.cli.temp.-000000-12345
-rw-r--r--    1 root     root     /tmp/jfrog.cli.temp.-111111-12345
-rw-r--r--    1 guest    users    /tmp/jfrog.cli.temp.-999999-12345

[Warn] failed while attempting to cleanup old CLI temp directories: Cannot remove path: /tmp/jfrog.cli.temp.-111111-12345 due to: operation not permitted

=== Files after jf -v ===
-rw-r--r--    1 root     root     /tmp/jfrog.cli.temp.-111111-12345
-rw-r--r--    1 guest    users    /tmp/jfrog.cli.temp.-999999-12345  <-- BUG: not deleted
```

**After Fix (local build):**
```
=== Files before jf -v ===
-rw-r--r--    1 guest    users    /tmp/jfrog.cli.temp.-000000-12345
-rw-r--r--    1 root     root     /tmp/jfrog.cli.temp.-111111-12345
-rw-r--r--    1 guest    users    /tmp/jfrog.cli.temp.-999999-12345

[Warn] failed while attempting to cleanup old CLI temp directories: failed to cleanup 1 file(s): /tmp/jfrog.cli.temp.-111111-12345 (operation not permitted)

=== Files after jf -v ===
-rw-r--r--    1 root     root     /tmp/jfrog.cli.temp.-111111-12345  <-- Only root's file remains
```

### DIFF
--- a/utils/io/fileutils/temp.go
+++ b/utils/io/fileutils/temp.go
@@ -84,6 +84,8 @@ func CleanOldDirs() error {
 		return errorutils.CheckError(err)
 	}
 	now := time.Now()
+	var failedDeletions []string
+
 	// Search for files/dirs that match the template.
 	for _, file := range files {
 		fileName := file.Name()
@@ -91,15 +93,28 @@ func CleanOldDirs() error {
 			var timestamp time.Time
 			timestamp, err = extractTimestamp(fileName)
 			if err != nil {
-				return errorutils.CheckErrorf("could not extract timestamp from file %s: %q", fileName, err)
+				// Collect timestamp extraction errors and continue
+				failedDeletions = append(failedDeletions,
+					fmt.Sprintf("%s (could not extract timestamp: %v)", fileName, err))
+				continue
 			}
 			// Delete old file/dirs.
 			if now.Sub(timestamp).Hours() > maxFileAge {
-				if err = RemovePath(path.Join(tempDirBase, fileName)); err != nil {
-					return err
+				filePath := path.Join(tempDirBase, fileName)
+				if err = RemovePath(filePath); err != nil {
+					// Collect deletion errors and continue processing
+					failedDeletions = append(failedDeletions,
+						fmt.Sprintf("%s (%v)", filePath, err))
+					continue
 				}
 			}
 		}
+	}
+
+	// At the end, return aggregated error message for all failures
+	if len(failedDeletions) > 0 {
+		return fmt.Errorf("failed to cleanup %d file(s): %s",
+			len(failedDeletions), strings.Join(failedDeletions, "; "))
 	}
 	return nil
 }
--- a/utils/io/fileutils/temp_test.go
+++ b/utils/io/fileutils/temp_test.go
@@ -2,6 +2,7 @@ package fileutils
 
 import (
 	"os"
+	"path"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


utils/io/fileutils/temp.go
```diff
@@ -84,23 +84,38 @@ func CleanOldDirs() error {
...
+					continue
 				}
 			}
 		}
 	}
+
+	// At the end, return aggregated error message for all failures
+	if len(failedDeletions) > 0 {
+		return fmt.Errorf("failed to cleanup %d file(s): %s",
+			len(failedDeletions), strings.Join(failedDeletions, "; "))
```
fluxxBot, line 117: let's wrap new errors in separate line.
strings.Join(failedDeletions, "; \n")

PR #1277, opened 2025-12-11
Fix: Handle Empty/Uninitialized .git Directory Gracefully

agrasth:
-----
# Fix: Handle Empty/Uninitialized .git Directory Gracefully

## Problem

When running the JFrog CLI upload command (`jf rt u`) with `--build-name` and `--build-number` flags, the CLI attempts to automatically collect Git build information.

If the current workspace contains a `.git` directory that is present but empty or uninitialized (specifically missing the `HEAD` or `config` files), the CLI crashes immediately with a "no such file or directory" error.

### Error Before Fix

```
[Error] open .git/HEAD: no such file or directory
[Error] open .git/HEAD: no such file or directory
[Error] open .git/HEAD: no such file or directory
```

The CLI assumed that if `.git` exists, it is a fully valid repository, and failed fatally when it cannot read the HEAD reference.

## Root Cause

In `utils/git.go`, the `getRevisionAndBranchPath()` and `readUrl()` functions directly attempt to open `.git/HEAD` and `.git/config` files without first checking if they exist:

```go
// Before fix - crashes if file doesn't exist
func (m *GitManager) getRevisionAndBranchPath() (revision, refUrl string, err error) {
    dotGitPath := filepath.Join(m.path, "HEAD")
    file, err := os.Open(dotGitPath)  // Crashes here!
        return
    }
    // ...
}
```

## Solution

Added existence checks before attempting to open `.git/HEAD` and `.git/config` files. If the files don't exist, the functions now log a warning and continue gracefully instead of returning an error:

```go
// After fix - checks existence first
func (m *GitManager) getRevisionAndBranchPath() (revision, refUrl string, err error) {
    dotGitPath := filepath.Join(m.path, "HEAD")

    // Check if HEAD file exists before trying to open it
    exists, err := fileutils.IsFileExists(dotGitPath, false)
    if err != nil {
        return "", "", err
    }
    if !exists {
        log.Warn("Git HEAD file not found at: " + dotGitPath + 
            ". The repository may be uninitialized or corrupt. Skipping VCS info collection.")
        return "", "", nil
    }
    // ... continue with normal processing
}
```

## Changes

**File**: `utils/git.go`

1. **`getRevisionAndBranchPath()`**: Added check for `.git/HEAD` existence before opening
2. **`readUrl()`**: Added check for `.git/config` existence before opening

### Unit Tests Added

Added three new test cases in `utils/git_test.go`:

1. **`TestReadConfigEmptyGitDir`**: Tests completely empty `.git` directory
2. **`TestReadConfigMissingHeadFile`**: Tests `.git` with config but no HEAD
3. **`TestReadConfigMissingConfigFile`**: Tests `.git` with HEAD but no config

**Before fix (jf 2.85.0):**
```bash
$ mkdir -p /tmp/test/.git  # Empty .git directory
$ cd /tmp/test
$ jf rt u "file.txt" "repo/" --build-name="test" --build-number=1
# Result: FAILURE
# Log: [Error] open /tmp/test/.git/HEAD: no such file or directory
```

**After fix:**
```bash
$ mkdir -p /tmp/test/.git  # Empty .git directory
$ cd /tmp/test
$ jf-fixed rt u "file.txt" "repo/" --build-name="test" --build-number=1
# Result: SUCCESS
# Log: [Warn] Git HEAD file not found at: /tmp/test/.git/HEAD. 
#      The repository may be uninitialized or corrupt. Skipping VCS info collection.
```

## Expected Behavior After Fix

The CLI now handles the missing HEAD/config files gracefully:
- Logs a warning that Git information could not be collected
- Ignores the Git context and proceeds with the artifact upload
- Provides a descriptive message indicating the repository may be uninitialized or corrupt

### DIFF
--- a/utils/git.go
+++ b/utils/git.go
@@ -119,6 +119,19 @@ func (m *GitManager) readUrl() {
 		return
 	}
 	dotGitPath := filepath.Join(m.path, "config")
+
+	// Check if config file exists before trying to open it.
+	// This handles cases where .git directory exists but is empty/uninitialized.
+	exists, err := fileutils.IsFileExists(dotGitPath, false)
+	if err != nil {
+		m.err = err
+		return
+	}
+	if !exists {
+		log.Debug("Git config file not found at: " + dotGitPath + ". Skipping URL collection.")
+		return
+	}
+
 	file, err := os.Open(dotGitPath)
 	if err != nil {
 		m.err = err
@@ -164,6 +177,18 @@ func (m *GitManager) readUrl() {
 
 func (m *GitManager) getRevisionAndBranchPath() (revision, refUrl string, err error) {
 	dotGitPath := filepath.Join(m.path, "HEAD")
+
+	// Check if HEAD file exists before trying to open it.
+	// This handles cases where .git directory exists but is empty/uninitialized.
+	exists, err := fileutils.IsFileExists(dotGitPath, false)
+	if err != nil {
+		return "", "", err
+	}
+	if !exists {
+		log.Warn("Git HEAD file not found at: " + dotGitPath + ". The repository may be uninitialized or corrupt. Skipping VCS info collection.")
+		return "", "", nil
+	}
+
 	file, err := os.Open(dotGitPath)
 	if errorutils.CheckError(err) != nil {
 		return
--- a/utils/git_test.go
+++ b/utils/git_test.go
@@ -1,14 +1,15 @@
 package utils
 
 import (
-	"github.com/jfrog/jfrog-client-go/utils/io/fileutils"
-	"github.com/stretchr/testify/assert"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"testing"
 	"time"
+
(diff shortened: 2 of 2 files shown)


utils/git.go
```diff
@@ -164,6 +177,18 @@ func (m *GitManager) readUrl() {
...
+
+	// Check if HEAD file exists before trying to open it.
+	// This handles cases where .git directory exists but is empty/uninitialized.
+	exists, err := fileutils.IsFileExists(dotGitPath, false)
+	if err != nil {
+		return "", "", err
+	}
+	if !exists {
+		log.Warn("Git HEAD file not found at: " + dotGitPath + ". The repository may be uninitialized or corrupt. Skipping VCS info collection.")
+		return "", "", nil
```
fluxxBot, line 189: why aren't we returning error here?

agrasth: The original bug was that it returned an error, which crashed the entire upload operation. The fix treats missing git files as a warning (logged) not an error (blocking)

PR #1294, opened 2026-01-05
Fix Windows temp file cleanup failures in build-add-dependencies command

bhanurp:
-----
## Description

The jf rt build-add-dependencies command intermittently fails on Windows with:

failed to close reader: remove <workspace>\jfrog-temp\jfrog.cli.temp.<...>:The process cannot access the file because it is being used by another process.
This occurs when antivirus or Windows indexing services temporarily lock temp files during cleanup.

## Solution

Fixed removeFile() to properly retry on Windows (10 retries × 500ms = 5s total wait time)
Updated Close() to log a warning instead of failing on Windows after retries exhausted (temp file cleanup is non-critical)
Non-Windows behavior remains unchanged (fail immediately on error).

### DIFF
--- a/utils/io/content/contentreader.go
+++ b/utils/io/content/contentreader.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"reflect"
+	"runtime"
 	"sort"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


itsmeleela (approved):
Looks good to me. In future we need to consider if we need to expose these time and retries config to user for windows

PR #1297, opened 2026-01-13
Add goroutine-local logger and BufferedLogger for parallel operations

eyalk007:
-----

- Add BufferedLogger that writes to io.Writer for isolated log capture
- Add goroutine-local logger support via sync.Map
- SetLoggerForGoroutine/ClearLoggerForGoroutine for per-goroutine logging
- GetLogger checks goroutine-local logger before falling back to global

### DIFF
--- a/utils/log/logger.go
+++ b/utils/log/logger.go
@@ -21,7 +21,7 @@ var goroutineLoggers sync.Map
 
 // goroutineLoggerCount tracks how many goroutine-local loggers are set.
 // Used as a fast-path check to avoid expensive getGoroutineID() calls
-// when no goroutine loggers exist (the common case).
+// when no goroutine loggers exist
 var goroutineLoggerCount int64
 
 // getGoroutineID extracts the current goroutine's ID from the runtime stack.


utils/log/bufferedlogger.go
```diff
@@ -0,0 +1,138 @@
...
+	if b.logLevel >= ERROR {
+		b.append(ERROR, a...)
+	}
+}
+
+func (b *BufferedLogger) Output(a ...interface{}) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	b.entries = append(b.entries, logEntry{level: -1, msg: fmt.Sprint(a...)})
+}
```
attiasas, line 74:
```suggestion
func (b *BufferedLogger) Output(a ...interface{}) {
	b.append(-1, a...)
}
```

 Output and logs are combined? if so, you can just call append here.. (make sure this is expected)

utils/log/bufferedlogger.go
```diff
@@ -0,0 +1,138 @@
...
+		case DEBUG:
+			target.Debug(entry.msg)
+		case INFO:
+			target.Info(entry.msg)
+		case WARN:
+			target.Warn(entry.msg)
+		case ERROR:
+			target.Error(entry.msg)
+		default:
+			target.Output(entry.msg)
```
attiasas, line 93: again, make sure this is expected

utils/log/logger.go
```diff
@@ -1,16 +1,44 @@
...
 	"runtime"
+	"strconv"
+	"sync"
+
+	"github.com/forPelevin/gomoji"
+	"github.com/gookit/color"
+	"golang.org/x/term"
 )
 
+// goroutineLoggers stores per-goroutine loggers for parallel scan isolation
```
attiasas:
```suggestion
// goroutineLoggers stores per-goroutine loggers for parallel isolated tasks
```

utils/log/logger.go
```diff
@@ -1,16 +1,44 @@
...
+	idField := bytes.Fields(buf[:n])[1]
+	id, _ := strconv.ParseInt(string(idField), 10, 64)
+	return id
+}
+
+// SetLoggerForGoroutine sets a logger for the current goroutine only.
+// Other goroutines will continue using the global logger.
+func SetLoggerForGoroutine(logger Log) {
+	goroutineLoggers.Store(getGoroutineID(), logger)
+}
```
attiasas, line 42: can't we just pass the id? why do we need to infer it from runtime?

attiasas: you can also use SetLogger if needed whenever you need to change

attiasas (commented):
Nice, please take a look at my commets, also add tests for the related changes

utils/log/logger.go
eyalk007: Passing the ID would require threading it through every function in the call chain - including internal HTTP clients, file utilities, and other low-level code that calls GetLogger() without knowing the context. The runtime extraction makes isolation automatic and transparent. SetLogger changes the global logger for ALL goroutines - we need per-goroutine isolation so parallel operations don't interfere with each other.

utils/log/bufferedlogger.go
eyalk007: Yes, this is expected. The default case handles two scenarios:
1. Output level logs (stored with level -1) - these are special logs that bypass level filtering
2. Any unexpected level values as a safety fallback

utils/log/logger.go
attiasas: yes but logger probably should be minimal as possiable since its methods are called all the time.
it could affect other places causing slowness

PR #1329, opened 2026-03-25
add build filter in pattern aql

reshmifrog:
-----
**Problem:**
When a spec has both build AND pattern, the old code didn't recognize it as a build search. It treated it as a pattern-only search (WILDCARD), which meant:

It generated an AQL query from the pattern — completely ignoring the build name
This AQL scanned the entire repository matching the pattern (could be millions of files)
Only after downloading all those results, it filtered them by the build's checksums in memory
For large repositories like the customer's docker-local-ash, this made the command run for an extremely long time or timeout.

Why Not Use a REST API Instead of AQL?

We explored this. Artifactory has several REST APIs:

Build Info API (GET /api/build/{name}/{number}) — returns build metadata, but artifact paths are relative to the module, not the full Artifactory path. You'd need extra lookups to find where each file actually lives.
Build Artifacts API (GET /api/builds/buildArtifacts/...) — returns file locations, but has indexing delays for recently-published builds (our CI tests caught this — it returned 0 results for a build that was just published).
Checksum Search (GET /api/search/checksum) — requires one HTTP call per artifact. For a build with 500 artifacts, that's 500 HTTP round-trips.
None of these REST APIs support combining a build filter with a pattern filter in a single call. We'd always need multiple API calls + client-side merging, which adds complexity and is slower than a single well-scoped AQL query.

**The Fix:**

We flipped the search strategy from "scan by pattern first, filter by build second" to "search by build first, filter by pattern second".

A build typically has 50–500 artifacts. A broad pattern can match millions of files. So searching the build first and then filtering by pattern is dramatically faster.

**What we changed:**

Routing — When the spec has both build and pattern, we now route it to the BUILD strategy instead of WILDCARD
Pattern filter — After the BUILD strategy fetches the build's artifacts (small set), we added a step that filters those results to only include files matching the pattern
AQL over dedicated API — When a pattern is present, we use build-scoped AQL (artifact.module.build.name = "X") instead of the dedicated build-artifacts REST API, because the REST API has indexing delays for recently-published builds

### DIFF
--- a/artifactory/services/utils/aqlquerybuilder.go
+++ b/artifactory/services/utils/aqlquerybuilder.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/jfrog/jfrog-client-go/utils"
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+	"github.com/jfrog/jfrog-client-go/utils/log"
 )
 
 const spaceEncoding = "%20"
@@ -382,13 +383,21 @@ func buildBuildNameQueryPart(params *CommonParams) string {
 	if params.Build == "" || params.ExcludeArtifacts || params.IncludeDeps {
 		return ""
 	}
-	buildName, _, err := ParseNameAndVersion(params.Build, true)
-	if err != nil || buildName == "" {
+	buildName, buildNumber, err := ParseNameAndVersion(params.Build, true)
+	if err != nil {
+		log.Debug("Failed to parse build name:", err.Error())
+		return ""
+	}
+	if buildName == "" {
 		return ""
 	}
 	// Unescape escaped slashes that are part of the spec identifier encoding.
 	buildName = strings.ReplaceAll(buildName, "\\/", "/")
-	return fmt.Sprintf(`"artifact.module.build.name":"%s",`, buildName)
+	query := fmt.Sprintf(`"artifact.module.build.name":"%s",`, buildName)
+	if buildNumber != "" && buildNumber != LatestBuildNumberKey {
+		query += fmt.Sprintf(`"artifact.module.build.number":"%s",`, buildNumber)
+	}
+	return query
 }
 
 func buildReleaseBundleQuery(params *CommonParams) (string, error) {
--- a/artifactory/services/utils/aqlquerybuilder.go
+++ b/artifactory/services/utils/aqlquerybuilder.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/jfrog/jfrog-client-go/utils"
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
-	"github.com/jfrog/jfrog-client-go/utils/log"
 )
 
 const spaceEncoding = "%20"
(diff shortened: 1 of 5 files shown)


artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -377,6 +378,19 @@ func buildExcludeQueryPart(params *CommonParams, useLocalPath, recursive bool) (
 	return excludeQuery, nil
 }
 
+func buildBuildNameQueryPart(params *CommonParams) string {
```
agrasth: Why is only the build name used in the filter, not the build version? When a user specifies "my-build/42", this filters by my-build but includes artifacts from all versions. Is this the intended behavior?

artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -377,6 +378,19 @@ func buildExcludeQueryPart(params *CommonParams, useLocalPath, recursive bool) (
...
+	if params.Build == "" || params.ExcludeArtifacts || params.IncludeDeps {
+		return ""
+	}
+	buildName, _, err := ParseNameAndVersion(params.Build, true)
```
agrasth: Should we consider logging the error from ParseNameAndVersion for debugging purposes, even if you return an empty string to gracefully continue. Something like `log.Debug("Failed to parse build name: %v", err)`

agrasth (approved):

artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -53,8 +54,9 @@ func CreateAqlBodyForSpecWithPattern(params *CommonParams) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	buildFilter := buildBuildNameQueryPart(params)
```
naveenku-jfrog: **buildBuildNameQueryPart** func name should also include build number in it, because func is taking both build name and build number in consideration while building query. Name Can be: buildBuildNameAndNumberQueryPart

naveenku-jfrog (changes requested):

naveenku-jfrog (approved):

artifactory/services/utils/searchutil.go
```diff
@@ -92,9 +93,97 @@ func SearchBySpecWithBuild(specFile *CommonParams, flags CommonConf) (readerCont
...
+	defer func() {
+		err = errors.Join(err, errorutils.CheckError(writer.Close()))
+	}()
+
+	pattern := prepareSourceSearchPattern(specFile.Pattern, specFile.Target)
+	repoPathFileTriples, _, err := createRepoPathFileTriples(pattern, specFile.Recursive)
+	if err != nil {
+		return nil, err
+	}
+
```
naveenku-jfrog, line 156: Below for loop is having O(n^3), can we optimise it. It will cause performance issue.

artifactory/services/utils/specutils.go
```diff
@@ -156,8 +156,20 @@ func (aql *Aql) UnmarshalJSON(value []byte) error {
 }
 
 func (params CommonParams) GetSpecType() (specType SpecType) {
+	hasNonTrivialPattern := params.Pattern != "" && params.Pattern != "*"
```
naveenku-jfrog, line 159: params.IncludeDeps condition can be added here only.

artifactory/services/utils/specutils.go
```diff
@@ -156,8 +156,20 @@ func (aql *Aql) UnmarshalJSON(value []byte) error {
...
 	switch {
-	case params.Build != "" && params.Aql.ItemsFind == "" && (params.Pattern == "*" || params.Pattern == ""):
+	case params.Build != "" && params.Aql.ItemsFind == "" && !(hasNonTrivialPattern && params.IncludeDeps):
+		// When a build is specified, use the BUILD path. The BUILD path uses the
```
naveenku-jfrog, line 167: Static failures.

naveenku-jfrog (changes requested):
Left some comments.

naveenku-jfrog (approved):

PR #1338, opened 2026-04-21
fix: build+pattern downloads against virtual repos

bhanurp:
-----

Problem:
  PR #1329 routed `--build + pattern` through a new path that fetched all
  build artifacts and post-filtered in Go using a string compare on `Repo`,
  which cannot resolve a virtual repo to its backing local — every result
  was dropped and downloads returned 0 files.

Solution:
  Move the pattern match into the AQL query itself via an AND of the build
  filter and the repo/path/name triples, letting Artifactory resolve virtual
  repos server-side. Remove the now-redundant client-side filter and its
  dead helpers from searchutil.go.

### DIFF
--- a/artifactory/services/utils/aqlquerybuilder.go
+++ b/artifactory/services/utils/aqlquerybuilder.go
@@ -14,6 +14,10 @@ import (
 
 const spaceEncoding = "%20"
 
+// errTransitiveMultiRepoMsg is shared by both pattern-based and build-with-pattern AQL
+// builders so the validation and its error text stay in lockstep.
+const errTransitiveMultiRepoMsg = "when searching or downloading with the transitive setting, the pattern must include a single repository only, meaning wildcards are allowed only after the first slash"
+
 var specialAqlCharacters = map[rune]string{
 	'/':  "%2F",
 	'\\': "%5C",
@@ -34,7 +38,7 @@ func CreateAqlBodyForSpecWithPattern(params *CommonParams) (string, error) {
 		return "", err
 	}
 	if params.Transitive && !singleRepo {
-		return "", errorutils.CheckErrorf("when searching or downloading with the transitive setting, the pattern must include a single repository only, meaning wildcards are allowed only after the first slash.")
+		return "", errorutils.CheckErrorf(errTransitiveMultiRepoMsg)
 	}
 	includeRoot := strings.Count(searchPattern, "/") < 2
 	triplesSize := len(repoPathFileTriples)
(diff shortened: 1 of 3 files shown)


artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -108,54 +112,98 @@ func handleArchiveSearch(triple RepoPathFile, archivePathFilePairs []RepoPathFil
 	return query
 }
 
+// createAqlBodyForBuildArtifacts builds the AQL body for build artifacts with no pattern or
+// exclusions. Safe wrapper: the underlying function only errors on user-supplied input, which
+// is nil here.
```
itsmeleela, line 117: Safe wrapper - yes that's currently true, but isn't it a fragile contract. If createAqlBodyForBuildArtifactsWithExclusions ever gets a new error path that fires on nil params, this method silently returns an empty body and triggers a confusing downstream failure. What do you think? I am just thinking out loud.

artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -108,54 +112,98 @@ func handleArchiveSearch(triple RepoPathFile, archivePathFilePairs []RepoPathFil
...
+	}
+	return fmt.Sprintf(`{%s%s}`, excludeQuery, buildOr), nil
+}
 
-	return `{` + excludeQuery + `"$or":[` + strings.Join(items, ",") + "]}"
+// buildPatternFilterForBuildArtifacts returns an AQL "$or" clause matching the pattern's
+// repo/path/name triples, or an empty string when no non-trivial pattern is supplied. Callers
+// compose it with the build-name/build-number filter under "$and".
+func buildPatternFilterForBuildArtifacts(params *CommonParams) (string, error) {
+	if params == nil || params.Pattern == "" || params.Pattern == "*" {
```
itsmeleela, line 150: What will happen if user passes ** as the value for pattern?

artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -108,54 +112,98 @@ func handleArchiveSearch(triple RepoPathFile, archivePathFilePairs []RepoPathFil
...
+		return "", err
 	}
 
-	// Build the exclusion query if params with exclusions are provided
-	excludeQuery := ""
-	if params != nil && len(params.GetExclusions()) > 0 {
-		var err error
-		excludeQuery, err = buildExcludeQueryPart(params, true, params.Recursive)
-		if err != nil {
-			excludeQuery = ""
```
itsmeleela, line 129: Earler it looks we are silently dropping. The new change is definiely the right thing to do, but it's a behavior change that could break callers who relied on the old silent-drop behavior. Isn't it?

artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -108,54 +112,98 @@ func handleArchiveSearch(triple RepoPathFile, archivePathFilePairs []RepoPathFil
...
+	}
+	return fmt.Sprintf(`"$or":[%s]`, handleRepoPathFileTriples(triples, nil, len(triples))), nil
 }
 
+// createAqlBodyForBuildDependencies builds the AQL body for build dependencies with no
+// exclusions. Safe wrapper: the underlying function only errors on user-supplied exclusions,
+// which are nil here.
 func createAqlBodyForBuildDependencies(builds []Build) string {
-	return createAqlBodyForBuildDependenciesWithExclusions(builds, nil)
+	body, _ := createAqlBodyForBuildDependenciesWithExclusions(builds, nil)
```
itsmeleela, line 171: If a user runs jf dl --build=<name> --spec with a pattern against a virtual repo, the dependency side will still have the bug isn't it? 

The reason I am asking this is: 

Here based on what I understand, the artifact path gets pattern injection via buildPatternFilterForBuildArtifacts. The dependency path (createAqlBodyForBuildDependenciesWithExclusions) does not , it only gets the exclusion refactor.

artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -108,54 +112,98 @@ func handleArchiveSearch(triple RepoPathFile, archivePathFilePairs []RepoPathFil
...
+	if err != nil {
+		return "", err
+	}
+	if params.Transitive && !singleRepo {
+		return "", errorutils.CheckErrorf(errTransitiveMultiRepoMsg)
+	}
+	if len(triples) == 0 {
+		return "", nil
+	}
+	return fmt.Sprintf(`"$or":[%s]`, handleRepoPathFileTriples(triples, nil, len(triples))), nil
```
itsmeleela, line 164: Where is this method? handleRepoPathFileTriples - we are sending nil as params. Are we sure this method doesnt have any code path inside it that assumes non-nil pairs, this will produce malformed AQL silently. This needs verification.

artifactory/services/utils/aqlquerybuilder.go
fluxxBot: I agree, it will download every dependency of build foo/1 to our local filesystem.

artifactory/services/utils/aqlquerybuilder.go
```diff
@@ -108,54 +112,98 @@ func handleArchiveSearch(triple RepoPathFile, archivePathFilePairs []RepoPathFil
...
-		excludeQuery, err = buildExcludeQueryPart(params, true, params.Recursive)
-		if err != nil {
-			excludeQuery = ""
-		}
+	patternOr, err := buildPatternFilterForBuildArtifacts(params)
+	if err != nil {
+		return "", err
 	}
+	if patternOr != "" {
+		return fmt.Sprintf(`{%s"$and":[{%s},{%s}]}`, excludeQuery, buildOr, patternOr), nil
```
fluxxBot, line 141: nit: can we have helper function for this? something like - `andClause(parts ...string)`

fluxxBot (changes requested):

artifactory/services/utils/aqlquerybuilder.go
bhanurp: Fair point. Switched the wrapper to return `(string, error)` and propagated through `createAqlQueryForBuild` → `fetchBuildArtifactsOrDependenciesSha1` and the `searchProps` call in `filterBuildArtifactsAndDependencies`. Same fix applied to the dependency-side `createAqlBodyForBuildDependencies` for symmetry. Updated the test to assert NoError on the nil-params wrapper.

artifactory/services/utils/aqlquerybuilder.go
bhanurp: Traced the behavior rather than assuming it: `**` as a standalone download pattern produces correct match-anything AQL today. The triple parser yields `repo="**"`, and `buildInnerQueryPart` already skips emitting a repo condition when the triple's repo is `*` or `**` (aqlquerybuilder.go:398), so path/name become `$match:"*"`/etc. — functionally equivalent to no pattern.

I initially added an `isTrivialPattern` helper to short-circuit `**` but reverted it — `**` isn't a documented standalone pattern for `jf rt dl` (the documented forms are `*`, `path/**/file` glob-style, or omit pattern), so adding special handling was YAGNI. The existing behavior is correct.

artifactory/services/utils/aqlquerybuilder.go
bhanurp: Yes, this is an intentional behavior change — flagged in the commit message/PR description. Before, a malformed `--exclusions` pattern silently produced a build-only AQL; now it errors. Guarding this with a subtest (`malformed exclusion propagates as error`) so it doesn't silently regress. Happy to add a CHANGELOG entry if the project uses one — didn't find one in the repo.

artifactory/services/utils/aqlquerybuilder.go
bhanurp: Checked this, the scenario `--build + pattern + --include-deps + virtual` routes to `WILDCARD`, not BUILD. From `specutils.go:GetSpecType`:

```go
patternRequiresWildcard := hasNonTrivialPattern && (params.IncludeDeps || params.Props != "" || params.ExcludeProps != "")
case params.Build != "" && params.Aql.ItemsFind == "" && !patternRequiresWildcard:
    specType = BUILD
```

So `IncludeDeps + non-trivial pattern` → WILDCARD. The WILDCARD path builds AQL with the pattern triples directly (including virtual-repo resolution server-side) and then filters by build, so deps are narrowed correctly.

This is already covered by the existing test at `aqlquerybuilder_test.go:377`:
```go
{
    name:     "build with IncludeDeps and pattern routes to WILDCARD",
    params:   CommonParams{Build: "my-build/1", Pattern: "repo-local/*", IncludeDeps: true},
    expected: WILDCARD,
},
```

Verified empirically against ecosys: `--include-deps + virtual repo + pattern` downloads correctly.

The dependency-side AQL does NOT need pattern injection because user traffic never hits the BUILD path with `IncludeDeps=true + non-trivial pattern`.

artifactory/services/utils/aqlquerybuilder.go
bhanurp: Verified by inspection. `handleRepoPathFileTriples` at aqlquerybuilder.go:78:

```go
func handleRepoPathFileTriples(pathFilePairs []RepoPathFile, archivePathFilePairs []RepoPathFile, pathPairSize int) string {
    var query string
    archivePathPairSize := len(archivePathFilePairs)  // len(nil) == 0
    for i := 0; i < pathPairSize; i++ {
        if archivePathPairSize > 0 {  // always false when we pass nil
            query += handleArchiveSearch(...)
        } else {
            query += buildInnerQueryPart(pathFilePairs[i])
        }
        ...
    }
    return query
}
```

Passing `nil` as `archivePathFilePairs` goes through `len(nil) == 0` → `archivePathPairSize == 0` → the `handleArchiveSearch` branch is never taken. `nil` is the intended way to opt out of archive-entry matching. It's also the shape used by callers that don't have archive entries to match on. Covered by `multi-triple pattern expands into $or` and end-to-end on ecosys (nested virtual, multi-backing virtual, aggregated-builds scenarios all exercise this path).

artifactory/services/utils/aqlquerybuilder.go
bhanurp: Replied on the parent thread with the routing trace and existing test reference — `IncludeDeps + pattern` is routed to `WILDCARD`, not BUILD, so deps never hit the BUILD-path helper without pattern narrowing in user flow. Let me know if you think the defense-in-depth variant (pattern filter on dep AQL too) is worth adding; I skipped it on YAGNI grounds since the routing invariant is already tested.

artifactory/services/utils/aqlquerybuilder.go
bhanurp: Reconsidered after checking scope: there are 3 `"$or":[%s]` sites in this diff, all within ~60 lines of each other in the same file. I tried extracting an `orClause` helper but reverted it — 3 call sites in one diff hop isn't real duplication worth a named abstraction, and the helper ends up being a 1-line `fmt.Sprintf` wrapper. Happy to revisit if another call site lands that would genuinely benefit. Leaving the 3 sites inlined.

itsmeleela (approved):

fluxxBot (approved):

PR #1339, opened 2026-04-23
fix(searchutil): nil-guard ResultItem.ToArtifact against nil receiver

SAY-5:
Addresses the downstream crash reported at jfrog/jfrog-cli#3450.

Multi-arch container push callers in jfrog-cli-artifactory (OCI `buildinfo.createPushBuildProperties`) reach `ToArtifact` on a nil `*ResultItem` when the manifest-digest SHA lookup misses in Artifactory. The direct `item.Name` deref crashes the build with:

```
panic: runtime error: invalid memory address or nil pointer dereference
artifactory/services/utils.(*ResultItem).ToArtifact(0x0) searchutil.go:580 +0x42
```

…and takes the Jenkins build down.

## Fix

Return a zero-value `buildinfo.Artifact` when `item` is nil so callers see an empty record and can treat it as "no matching artifact" upstream, rather than fatally panicking the process.

`go build ./artifactory/services/utils/...` clean.

### DIFF
--- a/artifactory/services/utils/searchutil.go
+++ b/artifactory/services/utils/searchutil.go
@@ -455,6 +455,16 @@ func (item ResultItem) GetItemRelativeLocation() string {
 }
 
 func (item *ResultItem) ToArtifact() buildinfo.Artifact {
+	// Multi-arch container push callers (jfrog-cli-artifactory OCI
+	// buildinfo.createPushBuildProperties) reach ToArtifact on the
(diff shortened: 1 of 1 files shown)


SAY-5:

SAY-5:
Could a maintainer approve the Frogbot scan environment for fork runs? The workflow needs that to start.

PR #1341, opened 2026-04-27
RTECO-1015 - Implement ListSkills method to fetch skills

udaykb2:
-----

### Overview:
- Implements ListSkills() method, which calls skills repository to fetch skills
- All validations are handled in jfrog-cli-artifactory & validated values are sent to this repo

### DIFF
--- a/artifactory/services/skills.go
+++ b/artifactory/services/skills.go
@@ -21,6 +21,9 @@ const (
 	SkillXrayStatusScanInProgress   = "SCAN_IN_PROGRESS"
 	SkillXrayStatusBlocked          = "BLOCKED"
 	SkillXrayStatusApproved         = "APPROVED"
+
+	SkillSortByUpdated   = "updated"
+	SkillSortByDownloads = "downloads"
 )
 
 // SkillXrayStatusResponse is the response from the Skills Xray gate status endpoint.
@@ -59,9 +62,9 @@ func (ss *SkillsService) ListVersions(repoKey, slug string) ([]SkillVersion, err
 
 func (ss *SkillsService) ListSkills(repoKey string, limit int, cursor, sortBy string) ([]SkillListItem, string, error) {
 	log.Debug(fmt.Sprintf("Listing skills in repo '%s'...", repoKey))
-	sort := "updated"
-	if sortBy == "downloads" {
-		sort = "downloads"
+	sort := SkillSortByUpdated
+	if sortBy == SkillSortByDownloads {
+		sort = sortBy
 	}
 	endpoint := fmt.Sprintf("skills?limit=%d&cursor=%s&sort=%s", limit, url.QueryEscape(cursor), sort)
 	body, err := ss.sendGet(repoKey, endpoint)
--- a/artifactory/services/skills.go
+++ b/artifactory/services/skills.go
@@ -76,19 +76,6 @@ func (ss *SkillsService) ListSkills(repoKey string, limit int, cursor, sortBy st
 	return wrapper.Items, wrapper.NextCursor, nil
 }
 
-func (ss *SkillsService) GetSkillDetail(repoKey, slug string) (*SkillDetail, error) {
-	log.Debug(fmt.Sprintf("Getting skill detail for '%s' in repo '%s'...", slug, repoKey))
-	body, err := ss.sendGet(repoKey, fmt.Sprintf("skills/%s", slug))
-	if err != nil {
-		return nil, err
-	}
-	var detail SkillDetail
-	if err = json.Unmarshal(body, &detail); err != nil {
-		return nil, errorutils.CheckErrorf("failed to parse skill detail response: %s", err.Error())
-	}
-	return &detail, nil
-}
-
 func (ss *SkillsService) SearchSkills(repoKey, query string, limit int) ([]SkillSearchResult, error) {
 	log.Debug(fmt.Sprintf("Searching skills in repo '%s' with query '%s'...", repoKey, query))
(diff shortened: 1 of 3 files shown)


artifactory/services/skills.go
```diff
@@ -57,6 +57,25 @@ func (ss *SkillsService) ListVersions(repoKey, slug string) ([]SkillVersion, err
 	return wrapper.Items, nil
 }
 
+func (ss *SkillsService) ListSkills(repoKey string, limit int, cursor, sortBy string) ([]SkillListItem, string, error) {
+	log.Debug(fmt.Sprintf("Listing skills in repo '%s'...", repoKey))
+	sort := "updated"
+	if sortBy == "downloads" {
+		sort = "downloads"
+	}
+	endpoint := fmt.Sprintf("skills?limit=%d&cursor=%s&sort=%s", limit, url.QueryEscape(cursor), sort)
```
itsmeleela, line 69: We are not validating the value of limit - isn't that needed? What will happen if I give 0 or negative numbers. may be add a test for the same and document clearly if needed

artifactory/services/skills.go
```diff
@@ -57,6 +57,25 @@ func (ss *SkillsService) ListVersions(repoKey, slug string) ([]SkillVersion, err
 	return wrapper.Items, nil
 }
 
+func (ss *SkillsService) ListSkills(repoKey string, limit int, cursor, sortBy string) ([]SkillListItem, string, error) {
+	log.Debug(fmt.Sprintf("Listing skills in repo '%s'...", repoKey))
+	sort := "updated"
```
itsmeleela: This silently ignores any invalid sortBy value (e.g. "name", "version") and falls through to "updated". This is fine defensively, but the behavior is invisible to the user, they pass "name", get "updated" results, and have no idea why. 

@bhanurp how are we doing in other places?

itsmeleela: Either validate strictly and return fmt.Errorf("unsupported sortBy value %q; supported: updated, downloads", sortBy), or add a comment explicitly documenting the silent-default contract for sortby

artifactory/services/skills.go
```diff
@@ -198,6 +217,21 @@ type SkillPropertySearchResult struct {
...
+	Summary       string        `json:"summary,omitempty"`
+	LatestVersion *SkillVersion `json:"latestVersion,omitempty"`
+	Versions      int           `json:"versions,omitempty"`
+	Updated       string        `json:"updated,omitempty"`
+}
+
+type skillListResponse struct {
+	Items      []SkillListItem `json:"items"`
+	NextCursor string          `json:"nextCursor,omitempty"`
+	Total      int             `json:"total,omitempty"`
```
itsmeleela, line 235: Total is parsed but not returned. If the CLI display will ever need "showing N of M", the interface signature will need a breaking change later. Intentional for now?

artifactory/services/skills.go
udaykb2: If we see other functions, we are not doing any validation here. Validations are happening, in jfrog-cli-artifactory only.

For example, if user gives other than sortBy "name", if we are not supporting "name", cli will throw error and tell him we are only supporting "updated" & "downloads".

artifactory/services/skills.go
udaykb2: These validations were handled in jfrog-cli-artifactory, if we see SearchSkills() which is below this funciton. There too we didn't check the values that we are passing, all of them were handled in jfrog-cli-artifactory, so we followed the same method here too

artifactory/services/skills.go
udaykb2: Can you please elaborate your question?

we, here simply followed the structure mentioned in https://docs.jfrog.com/artifactory/docs/skills-repositories#:~:text=%7B%0A%20%20%22items%22,%3A%202%0A%7D

artifactory/services/skills.go
```diff
@@ -57,6 +57,25 @@ func (ss *SkillsService) ListVersions(repoKey, slug string) ([]SkillVersion, err
 	return wrapper.Items, nil
 }
 
+func (ss *SkillsService) ListSkills(repoKey string, limit int, cursor, sortBy string) ([]SkillListItem, string, error) {
+	log.Debug(fmt.Sprintf("Listing skills in repo '%s'...", repoKey))
```
naveenku-jfrog, line 64: There is no need of using fmt.Sprintf statements here. log.Debug itself provide these functionality.

artifactory/services/skills.go
naveenku-jfrog: Why we are not defining this string in constant?

artifactory/services/skills.go
```diff
@@ -57,6 +57,25 @@ func (ss *SkillsService) ListVersions(repoKey, slug string) ([]SkillVersion, err
...
+	sort := "updated"
+	if sortBy == "downloads" {
+		sort = "downloads"
```
naveenku-jfrog: Why we are not defining this string in constant?

naveenku-jfrog (changes requested):

artifactory/services/skills.go
udaykb2: We are following this approach in order to align with the way log.Debug written in other functions.

Please refer ListVersions(), SearchSkills(), we are using the same log.Debug+Sprintf.

artifactory/services/skills.go
udaykb2: done, changed applied

artifactory/services/skills.go
udaykb2: updated code

bhanurp (approved):

naveenku-jfrog (approved):

PR #1347, opened 2026-05-20
Fix race in ContentReader causing 'send on closed channel' panic

fluxxBot:
-----

## Summary                                                                                                                                                                           
                                                                                                                                                                                     
  Fixes `panic: send on closed channel` in `ContentReader` triggered intermittently by concurrent use of `NextRecord` and `Reset` (or by callers that invoke `Reset` internally, e.g.  
  `Length`, `SortContentReader`).
                                                                                                                                                                                       
  Resolves [RTECO-1263](https://jfrog-int.atlassian.net/browse/RTECO-1263).                                                                                                            
                                      
  ## Background                                                                                                                                                                        
                                                                                                                                                                                     
  `ContentReader.NextRecord` spawns a producer goroutine once (via `sync.Once`) that streams JSON array elements from one or more files into `cr.dataChannel`. The receiver in         
  `NextRecord` and the producer in `readSingleFile` both read `cr.dataChannel` from the struct field on each access.
                                                                                                                                                                                       
  `Reset()` replaces both `cr.dataChannel` and `cr.once` without any synchronization. This created three independent race conditions that together produce the panic:                  
                                      
  1. **Producer / Reset race** — a producer goroutine paused between sends (between files, or while blocked on a full buffer) could resume after `Reset()` had swapped the channel     
  field. The producer's next send would land on the new channel, which could already have been closed by the next cycle's `defer close`.                                             
  2. **Receiver / Reset race** — `NextRecord` could fire `once.Do` against the old `once`, then read `cr.dataChannel` *after* `Reset()` had swapped both fields, causing the receiver  
  and producer to operate on different channels (deadlock).                                                                                                                            
  3. **Defer / send mismatch** — `defer close(cr.dataChannel)` evaluates the channel argument at defer-registration time, but `cr.dataChannel <- item` re-reads the field on every
  send. The defer and the send could refer to two different channels.                                                                                                                  
                                                                                                                                                                                     
  The race is timing-dependent; it surfaces in CI under load (e.g. `TestPushFatManifestImageWithNestedPath` in the docker test suite) but is hard to reproduce locally without `-race` 
  and concurrent stress.                                                                                                                                                             
                                                                                                                                                                                       
  Stack trace from the failing CI run:                                                                                                                                               
                                      
  panic: send on closed channel
  goroutine 2699 [running]:
  github.com/jfrog/jfrog-client-go/utils/io/content.(*ContentReader).readSingleFile
      .../utils/io/content/contentreader.go:216 +0x492
  github.com/jfrog/jfrog-client-go/utils/io/content.(*ContentReader).run(...)
      .../utils/io/content/contentreader.go:178
  github.com/jfrog/jfrog-client-go/utils/io/content.(*ContentReader).NextRecord.func1.1()
      .../utils/io/content/contentreader.go:78 +0x7a

### DIFF
--- a/utils/io/content/contentreader_test.go
+++ b/utils/io/content/contentreader_test.go
@@ -191,7 +191,7 @@ func closeAndAssert(t *testing.T, reader *ContentReader) {
 func TestContentReaderResetRace(t *testing.T) {
 	tmp, err := os.CreateTemp("", "aql-race-*.json")
 	assert.NoError(t, err)
-	defer os.Remove(tmp.Name())
+	defer func() { _ = os.Remove(tmp.Name()) }()
 	payload := map[string]interface{}{
 		"results": []map[string]string{
 			{"repo": "r1"}, {"repo": "r2"}, {"repo": "r3"},
--- a/utils/io/content/contentreader_test.go
+++ b/utils/io/content/contentreader_test.go
@@ -1,10 +1,12 @@
 package content
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/jfrog/jfrog-client-go/utils/io/fileutils"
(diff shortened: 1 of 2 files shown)


utils/io/content/contentreader_test.go
```diff
@@ -183,6 +185,42 @@ func closeAndAssert(t *testing.T, reader *ContentReader) {
 	assert.NoError(t, reader.Close(), "Couldn't close reader")
 }
 
+// Regression test for "panic: send on closed channel" caused by Reset() racing
+// with an in-flight producer goroutine. Without the fix (local channel snapshot
+// in NextRecord + WaitGroup in Reset), this test panics under -race.
+func TestContentReaderResetRace(t *testing.T) {
+	tmp, err := os.CreateTemp("", "aql-race-*.json")
+	assert.NoError(t, err)
+	defer os.Remove(tmp.Name())
```
naveenku-jfrog: errcheck is already failing CI here. Change this to:

defer func() { _ = os.Remove(tmp.Name()) }()
or if you want it asserted in the test:


defer func() { assert.NoError(t, os.Remove(tmp.Name())) }()

naveenku-jfrog (changes requested):

naveenku-jfrog (approved):

PR #1351, opened 2026-06-03
Use Windows system certificate store for CLI TLS transport

RemiBou:
## Summary

- When `x509.SystemCertPool()` returns a nil pool (notably on Windows), fall back to `x509.NewCertPool()` so TLS transport setup never passes a nil `RootCAs` pool.
- Consolidate platform-specific `loadSystemRoots` helpers into a single implementation using `SystemCertPool`.
- Remove the obsolete comment in `loader.go` and add unit tests for cert pool loading and transport configuration.

Based on [Cesar-DC/jfrog-client-go@bugfix-issue-3141](https://github.com/Cesar-DC/jfrog-client-go/tree/bugfix-issue-3141), discussed in [jfrog/jfrog-cli#3441](https://github.com/jfrog/jfrog-cli/pull/3441). Related to [jfrog/jfrog-cli#2309](https://github.com/jfrog/jfrog-cli/issues/2309).

## Test plan

- [x] `go test ./auth/cert/...`
- [x] Verify TLS connectivity on Windows against a host using the OS trust store

Made with [Cursor](https://cursor.com)


auth/cert/sslutils_windows.go
```diff
@@ -0,0 +1,13 @@
+package cert
+
+import (
+	"crypto/x509"
+)
+
+func loadSystemRoots() (*x509.CertPool, error) {
+	pool, err := x509.SystemCertPool()
```
vitaliil-jfrog, line 11: Would it make sense to return `err` early if there is one?

vitaliil-jfrog (approved):

cyrilc-pro (approved):

PR #1358, opened 2026-07-20
XRAY-156161 - Add optional workspace for config profile obtaining

VidrevichMichael:
-----

### DIFF
--- a/xray/services/xsc/xsc.go
+++ b/xray/services/xsc/xsc.go
@@ -74,6 +74,13 @@ func (xs *XscInnerService) GetConfigProfileByUrl(repoUrl string) (*services.Conf
 	return configProfileService.GetConfigurationProfileByUrl(repoUrl)
 }
 
+func (xs *XscInnerService) GetConfigProfileByUrlAndWorkspace(repoUrl, workspaceName string) (*services.ConfigProfile, error) {
+	configProfileService := services.NewConfigurationProfileService(xs.client)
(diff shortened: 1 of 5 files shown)


VidrevichMichael:

xray/services/xsc/xsc.go
```diff
@@ -74,6 +74,13 @@ func (xs *XscInnerService) GetConfigProfileByUrl(repoUrl string) (*services.Conf
 	return configProfileService.GetConfigurationProfileByUrl(repoUrl)
 }
 
+func (xs *XscInnerService) GetConfigProfileByUrlAndWorkspace(repoUrl, workspaceName string) (*services.ConfigProfile, error) {
+	configProfileService := services.NewConfigurationProfileService(xs.client)
+	configProfileService.XrayDetails = xs.XrayDetails
+	configProfileService.ScopeProjectKey = xs.ScopeProjectKey
+	return configProfileService.GetConfigurationProfileByUrlAndWorkspace(repoUrl, workspaceName)
+}
```
attiasas, line 82: I don't know if we should create a new func or adjusting the old by adding the new param to it.

It will be easier to maintain that way and adjusted in all used places so they can, if they want to pass this new param.

Your call

attiasas (approved):
Nice! LGTM.

This new option is only avaliable from Xray minimum version, we should consider where to check and how to react if user uses CLI with Xray that does not meet the required version. (it should be validated here or/and when passing this. option in the CLI)

PR #1372, opened 2026-08-05
adding fields to XscAnalyticsGeneralEventFinalize

eranturgeman:
-----

### DIFF
--- a/xsc/services/analytics.go
+++ b/xsc/services/analytics.go
@@ -250,8 +250,11 @@ type GitIntegrationEvent struct {
 
 type XscAnalyticsGeneralEventFinalize struct {
 	XscAnalyticsBasicGeneralEvent
-	MultiScanId string `json:"multi_scan_id,omitempty"`
-	GitRepoUrl  string `json:"git_repository,omitempty"`
+	MultiScanId           string   `json:"multi_scan_id,omitempty"`
+	GitRepoUrl            string   `json:"git_repository,omitempty"`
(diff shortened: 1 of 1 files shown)


xsc/services/analytics.go
```diff
@@ -250,8 +250,11 @@ type GitIntegrationEvent struct {
 
 type XscAnalyticsGeneralEventFinalize struct {
 	XscAnalyticsBasicGeneralEvent
-	MultiScanId string `json:"multi_scan_id,omitempty"`
-	GitRepoUrl  string `json:"git_repository,omitempty"`
+	MultiScanId          string   `json:"multi_scan_id,omitempty"`
+	GitRepoUrl           string   `json:"git_repository,omitempty"`
+	ScanTypesExecuted    []string `json:"scan_types_executed,omitempty"`
+	ProjectKey           string   `json:"project_key,omitempty"`
```
attiasas, line 256: This could be reported at `XscAnalyticsBasicGeneralEvent`. known from the start

attiasas (approved):
LGTM, see my comment

xsc/services/analytics.go
eranturgeman: correct but this is not something we need to keep in the DB (no column and not required). when enter the finalize event we need to pull the start event from the DB and I dont think it is necessary to do a migration and adding a column and change query just for saving this at the beginning with no actual need

PR #1381, opened 2026-08-25
RTDEV-97165 - Add limit/cursor params to ListVersions

udaykb2:
The Skills API versions endpoint (/api/v1/skills/{slug}/versions) does not support a sort parameter at all, confirmed by decompiling the server's SkillsResource.listSkillVersions bytecode — it has only 3 parameters: slug, limit, cursor. Remove the dead sortBy parameter and its callers.

ListVersions now takes (repoKey, slug string, limit int, cursor string), returning ([]SkillVersion, nextCursor string, error).

Mirrors ListSkills' pagination pattern for the versions endpoint using only the confirmed parameters (limit, cursor). The returned nextCursor is the next page's cursor, or "" when the current page is the last one (the server omits nextCursor from the JSON entirely in that case).

limit/cursor behavior verified empirically against a live Artifactory instance (agent-skills/readme-standard on bukgradlefix.jfrogdev.org): a short limit truncates the result and returns a real nextCursor; passing that cursor back resumes correctly after the last item.

Code review fixes from decompiling the actual server:
  - slug is url.PathEscape'd before being interpolated into request path
  - VersionExists now paginates across all pages to avoid false negatives on skills with unusually large version histories
  - Pagination loop stops if server returns non-advancing cursor

Adds comprehensive unit test coverage via httptest mock server.

-----

### DIFF
--- a/artifactory/emptymanager.go
+++ b/artifactory/emptymanager.go
@@ -118,6 +118,7 @@ type ArtifactoryServicesManager interface {
 	ListSkills(repoKey string, limit int, cursor, sortBy string) ([]services.SkillListItem, string, error)
 	SearchSkills(repoKey, query string, limit int) ([]services.SkillSearchResult, error)
 	SkillVersionExists(repoKey, slug, version string) (bool, error)
+	SkillExists(repoKey, slug string) (bool, error)
 	SearchSkillsByProperty(query string) ([]services.SkillPropertySearchResult, error)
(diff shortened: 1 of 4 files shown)


udaykb2:

fluxxBot (approved):

udaykb2:
recheck

udaykb2:

PR #1383, opened 2026-08-26
XRAY-156994 - Add Scan Cdx upload endpoint

eranturgeman:
-----

### DIFF
--- a/README.md
+++ b/README.md
@@ -2794,7 +2794,7 @@ version, err := xscManager.GetVersion()
 Upload a CycloneDx SBOM directly to Xray for scanning, without requiring an Artifactory upload.
 
 ```go
-resp, err := xscManager.UploadScanCdx(services.UploadScanCdxParams{
+resp, err := xrayManager.Xsc().UploadScanCdx(services.UploadScanCdxParams{
     RepoName: "my-repo",
     RepoPath: "path/in/repo",
     FileName: "results.cdx.json",
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@
       - [Creating New XSC Service Manager](#creating-new-xsc-service-manager)
       - [Using XSC Services](#using-xsc-services)
       - [Fetching XSC's Version](#fetching-xscs-version)
+      - [Upload Scan CDX](#upload-scan-cdx)
       - [Report XSC analytics metrics](#report-xsc-analytics-metrics)
         - [Add Analytics General Event](#add-analytics-general-event)
         - [Update Analytics General Event](#update-analytics-general-event)
@@ -2755,6 +2756,19 @@ xscManager, err := xsc.New(serviceConfig)
 
 ```go
 version, err := xscManager.GetVersion()
+```
+
+#### Upload Scan CDX
+
+Upload a CycloneDx SBOM directly to Xray for scanning, without requiring an Artifactory upload.
+
+```go
+resp, err := xscManager.UploadScanCdx(services.UploadScanCdxParams{
+    RepoName: "my-repo",
+    RepoPath: "path/in/repo",
+    FileName: "results.cdx.json",
+    Bom:      bomJson,
+})
 ```
 
 #### Report XSC Analytics Metrics
--- a/http/httpclient/clientBuilder.go
+++ b/http/httpclient/clientBuilder.go
@@ -122,7 +122,6 @@ func (builder *httpClientBuilder) createDefaultHttpTransport() *http.Transport {
 		DialContext: (&net.Dialer{
 			Timeout:   builder.dialTimeout,
 			KeepAlive: 20 * time.Second,
-			DualStack: true,
 		}).DialContext,
 		MaxIdleConns:          100,
 		IdleConnTimeout:       90 * time.Second,
--- a/xsc/services/uploadscancdx.go
+++ b/xsc/services/uploadscancdx.go
@@ -21,10 +21,10 @@ type UploadScanCdxService struct {
 }
 
 type UploadScanCdxParams struct {
-	RepoName string          `json:"repo_name"`
-	RepoPath string          `json:"repo_path"`
-	FileName string          `json:"file_name"`
-	Bom      json.RawMessage `json:"bom"`
+	RepoName string `json:"repo_name"`
+	RepoPath string `json:"repo_path"`
+	FileName string `json:"file_name"`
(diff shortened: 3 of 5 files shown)


http/httpclient/clientBuilder.go
```diff
@@ -122,7 +122,6 @@ func (builder *httpClientBuilder) createDefaultHttpTransport() *http.Transport {
 		DialContext: (&net.Dialer{
 			Timeout:   builder.dialTimeout,
 			KeepAlive: 20 * time.Second,
-			DualStack: true,
```
eranturgeman: **For Reviewer:**
- What the fix is: removed DualStack: true from the net.Dialer literal in clientBuilder.go. That field has been a no-op since Go 1.12 (Go does fast IPv4/IPv6 fallback automatically now), so staticcheck's SA1019 rule flags setting it as deprecated/dead.
  - Why we had to add it: not because our code needed it — the CI's "Static Code Analysis" step runs golangci-lint over the entire repo, not just the diff. Since this pre-existing line failed that check, the whole job goes red regardless of what the PR actually changed.
  - How it's related to our change: it isn't. git log/git diff confirm that file was last touched by an unrelated commit (#887) and isn't part of our diff at all. It's unrelated tech debt that surfaced now — most likely a golangci-lint/staticcheck version bump in the CI
  action started enforcing (or newly including) this rule. Any PR against this repo right now would hit the same failure.

README.md
```diff
@@ -2788,6 +2789,19 @@ xscManager, err := xsc.New(serviceConfig)
 version, err := xscManager.GetVersion()
 ```
 
+#### Upload Scan CDX
+
+Upload a CycloneDx SBOM directly to Xray for scanning, without requiring an Artifactory upload.
+
+```go
+resp, err := xscManager.UploadScanCdx(services.UploadScanCdxParams{
```
Jordanh1996: snippet calls `xscManager.UploadScanCdx(...)` on a manager from `xsc.New(serviceConfig)`(from line 2780-2781 in this README), but XscServicesManager has no such method - the `UploadScanCdx` method is through the inner `Xsc` method

xsc/services/uploadscancdx.go
```diff
@@ -0,0 +1,67 @@
...
+	client          *jfroghttpclient.JfrogHttpClient
+	XrayDetails     auth.ServiceDetails
+	ScopeProjectKey string
+}
+
+type UploadScanCdxParams struct {
+	RepoName string `json:"repo_name"`
+	RepoPath string `json:"repo_path"`
+	FileName string `json:"file_name"`
+	Bom      string `json:"bom"`
```
Jordanh1996, line 28: Bom string + the server's html.UnescapeString corrupts real BOMs

The server (xray #22107, api_scan_cdx_upload.go) runs req.Bom = html.UnescapeString(req.Bom) right after deserializing, before verifyCdxIsValid. That unescape reverses nothing: json.Marshal escapes &/</> as \u0026/\u003c/\u003e, and json.Unmarshal already restores them — I verified the decoded Bom is byte-identical to what the client sent. So the call only rewrites &-sequences the BOM author actually wrote:

- ?id=1&copy=2&reg=3 → ?id=1©=2®=3 (Go resolves legacy named entities without a semicolon) — upload succeeds, corrupted URL gets stored and indexed
- "description":"Tom &amp; Jerry utils" → Tom & Jerry utils — same, silent
- "value":"password = &quot;hunter2&quot;" → injects a raw " into the JSON string → 400 bom is not a valid CycloneDX document. processScanResults turns that into AddGeneralError, so the whole audit/Frogbot scan fails on a file that uploads fine via Artifactory today.

Fix: make the field Bom json.RawMessage on both ends and delete the server's UnescapeString. That removes the round-trip entirely, makes reintroducing the unescape impossible, and drops the wire size (the BOM is currently indented and JSON-string-escaped — 1.61x vs compact, against the server's 100MB cap). If the type has to stay a string, the server call still needs to go.

xsc/services/uploadscancdx.go
```diff
@@ -0,0 +1,67 @@
...
+	Repository string `json:"repository"`
+	Path       string `json:"path"`
+}
+
+func NewUploadScanCdxService(client *jfroghttpclient.JfrogHttpClient) *UploadScanCdxService {
+	return &UploadScanCdxService{client: client}
+}
+
+func (us *UploadScanCdxService) getUploadScanCdxURL() string {
+	return utils.AppendScopedProjectKeyParam(utils.AddTrailingSlashIfNeeded(us.XrayDetails.GetUrl())+xscutils.XscInXraySuffix+uploadScanCdxAPIUrl, us.ScopeProjectKey)
```
Jordanh1996: optional:
I know this is this project convention. but this sort of concatenation is many times error prone. many languages has standard safe functions to construct paths using partials. including go's "url.JoinPath"

xsc/services/uploadscancdx.go
eranturgeman: I tested this directly rather than reasoning about it — ran your three examples through the actual html.EscapeString → html.UnescapeString pair used here:

  ?id=1&copy=2&reg=3                       → round-trips exactly, no © or ®
  Tom &amp; Jerry utils (literal text)      → round-trips exactly, no silent unescape
  {"description":"password = &quot;hunter2&quot;"} → round-trips exactly, no quote injection

  All three come back byte-identical. The reason: escapeStructStrings (api_utils.go) runs html.EscapeString on req.Bom right after JSON decode and before the html.UnescapeString call you flagged — so the unescape isn't operating on raw, never-escaped BOM text (which is
  what your examples assume and where the corruption would occur); it's undoing the escape that just ran seconds earlier. EscapeString always emits the semicolon-terminated form (&amp;, not &amp), so UnescapeString consumes exactly that token and stops, never re-parsing
  the trailing text as a new entity. So this isn't a bug — no change needed on either side.

README.md
eranturgeman: updated

xsc/services/uploadscancdx.go
eranturgeman: done

Jordanh1996 (approved):

xsc/services/uploadscancdx.go
eranturgeman: @Jordanh1996 I also added a test in xray to verify we do not corrupt the BOM and its indentations

PR #1386, opened 2026-08-31
artifactory: skip download when local file matches Sha256

### LINKED ISSUE #1384: DownloadFiles downloads files already on disk when Sha256 and Size are given
### Describe the bug

When downloading files, the DownloadService skips downloading if the file already exists on disk. It does this by comparing MD5 and SHA1 from the parameters to files on disk within downloadFileIfNeeded. However, if Sha256 and Size are provided in the parameters, we skip fetching info from Artifactory. As a consequence, the MD5 and SHA1 are unknown, and the client has no way to know if the file on disk matches what we need.

`fileutils.IsEqualToLocalFile` already calculates the SHA256 sum, but it's an exported function that only takes md5 and sha1 args. Maybe a new function is in order.

### Current behavior

(*DownloadService).DownloadFiles always redownloads when Sha256 and Size fields are set.

### Reproduction steps

```go
param := services.NewDownloadParams()
param.Sha256 = sha256sum
param.Size = size

// etc...

// Always downloads the file, even if already present on disk at destination.
rtMgr.DownloadFiles(param)
```

### Expected behavior

(*DownloadService).DownloadFiles avoid re-downloading files when Sha256 and Size fields are set.

### JFrog Client-Go version

v1.55.0

### JFrog CLI version (if applicable)

_No response_

### Operating system type and version

Ubuntu 26.04

### JFrog Artifactory version

_No response_

### JFrog Xray version

_No response_

amarkdotdev:
Fixes #1384

DownloadFiles now compares local files by Sha256 (and size when available) when Md5/Sha1 are not known, so downloads that bypass AQL no longer re-fetch existing files.

## Test plan
- [x] go test ./utils/io/fileutils/... -count=1 -run TestIsEqualToLocalFile (pass)

### DIFF
--- a/artifactory/services/download.go
+++ b/artifactory/services/download.go
@@ -594,6 +594,12 @@ func (ds *DownloadService) downloadFileIfNeeded(downloadPath, localPath, localFi
 	if err != nil {
 		return err
 	}
+	if !isEqual && downloadData.Dependency.Sha256 != "" {
+		isEqual, err = fileutils.IsEqualToLocalFileBySha256(localFilePath, downloadData.Dependency.Sha256, downloadData.Dependency.Size)
+		if err != nil {
+			return err
+		}
+	}
 	if isEqual {
 		log.Debug(logMsgPrefix+"File already exists locally:", localFilePath)
 		if ds.Progress != nil {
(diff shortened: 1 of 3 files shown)


amarkdotdev:
Friendly ping for maintainer eyes on this whenever convenient.

PR #1389, opened 2026-09-10
Support streaming POST requests

asafgabai:
-----


SaharBracha (approved):

http/httpclient/client.go
```diff
@@ -109,6 +109,35 @@ func (jc *HttpClient) SendPut(url string, content []byte, httpClientsDetails htt
...
+	req.Close = true
+
+	copyHeaders(httpClientsDetails, req)
+	setAuthentication(req, httpClientsDetails)
+	addUserAgentHeader(req)
+	addUberTraceIdHeaderIfSet(req)
+
+	log.Debug(fmt.Sprintf("Sending HTTP POST request to: %s", url))
+	resp, err = jc.client.Do(req) // #nosec G704 -- CLI/library; URL from user/config, runs in user environment
+	if errorutils.CheckError(err) != nil || resp == nil {
```
RemiBou, line 128: need to consume whole body when there are errors

RemiBou (approved):

http/httpclient/client.go
asafgabai: In case of HTTP errors (4xx/5xx) `Do` succeeds and the whole body is read as wanted. In case of error from the `Do` function, usually the response returned is nil - the convention in all other existing requests in this scenario is not to consume the response, but to throw the returned error.

PR #1391, opened 2026-09-15
XRAY-160563 - Add service to get transitive data from Catalog

VidrevichMichael:
-----


catalog/services/transitivecontextual.go
Jordanh1996, line 1: I'd like to argue that "transitive" is the application usecase. But the endpoint has no mention of anything being transitive. so as a pure wrapper I would omit any reference of the application which is an outcome of the capability

catalog/services/transitivecontextual.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/jfrog/jfrog-client-go/auth"
+	"github.com/jfrog/jfrog-client-go/http/jfroghttpclient"
+	"github.com/jfrog/jfrog-client-go/utils"
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+)
+
+const transitiveContextualApi = "api/v1/dependencies/contextual"
```
attiasas: is this API new? should we check minimal version of Catalog?

catalog/services/transitivecontextual.go
```diff
@@ -0,0 +1,82 @@
...
+
+// PackageVersionKey identifies a package by type, name, namespace, version and ecosystem.
+type PackageVersionKey struct {
+	Type      string `json:"type"`
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+	Version   string `json:"version"`
+	Ecosystem string `json:"ecosystem"`
+}
```
attiasas: `PackageVersionKey` Already exists at `xray/services/utils/remediation`
Let's try to see how to reuse it and not define duplicate structs

attiasas (commented):
Nice job! please take a look at my comments.

In addition, I agree that `transitive` does not included at the API route or in any structs, lets try to rename more fit to this API, maybe somthing with `dependencies` and/or `contexual`

Consider adding actual integration tests for this at `tests` and try to use tests fixtures

catalog/services/transitivecontextual.go
VidrevichMichael: no, it's the same we use to get vulnerability paths for JAS binary

Jordanh1996 (approved):

catalog/services/indirectcontextual.go
```diff
@@ -0,0 +1,74 @@
...
+	CatalogDetails  auth.ServiceDetails
+	ScopeProjectKey string
+}
+
+func NewIndirectContextualService(client *jfroghttpclient.JfrogHttpClient) *IndirectContextualService {
+	return &IndirectContextualService{client: client}
+}
+
+func (tc *IndirectContextualService) getUrl() string {
+	return utils.AppendScopedProjectKeyParam(tc.CatalogDetails.GetUrl()+indirectContextualApi, tc.ScopeProjectKey)
```
Jordanh1996, line 47: consider using go's `url.JoinPath`

VidrevichMichael: `JoinPath` does not support url parameters when `AppendScopedProjectKeyParam` does

